PR #109, opened 2015-01-15
Fixed empty return in top_meta_image

ecesena:
Return empty unicode in ContentExtractor.top_meta_image() when no image is found. It was returning the article's url.

Example:
http://www.usnews.com/news/entertainment/articles/2015/01/15/boyhood-keaton-moore-are-oscar-favorites-in-vegas
(note that USNews often updates its articles, e.g. by adding an image -- at the time of writing it was without images)

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -316,7 +316,9 @@ class ContentExtractor(object):
 
         top_meta_image = try_one or try_two or try_three or try_four
 
-        return urllib.parse.urljoin(article_url, top_meta_image)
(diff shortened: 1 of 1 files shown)


codelucas:
Nicely done, my only comment is that newspaper on `master` is a python 3 library, so `return u''` should be `return ''`.

Don't worry about it though, I'll merge this on the command line and apply that tiny fix before the code goes in later today.

ecesena:
Ops, sorry about that and thank you!

PR #110, opened 2015-01-15
Fixed nodeToString() to return valid HTML

ecesena:
In some cases the output of nodeToString() is not valid HTML, e.g. for iframe nodes.

Solution found here: http://stackoverflow.com/questions/27020950/lxml-modify-tags-prevent

An example of bad behavior is an article with a youtube video inside. The video embed code is transformed into `<iframe ... />`, that is not valid, and thus the result in the browser is a page with no visible content after the video.

### DIFF
--- a/newspaper/parsers.py
+++ b/newspaper/parsers.py
@@ -85,7 +85,7 @@ class Parser(object):
         """`decode` is needed at the end because `etree.tostring`
         returns a python bytestring
         """
-        return lxml.etree.tostring(node).decode()
+        return lxml.etree.tostring(node, method='html').decode()
 
     @classmethod
     def replaceTag(cls, node, tag):


codelucas:
Testing this locally right now, thanks for another good PR @ecesena!

codelucas:
@ecesena Your work is good, it's gonna be a little longer until I merge, currently trying to fix some things on the python-2 branch.

ecesena:
Thank you, no rush on my side - I'll pull req something more complex for the title extraction by the end of the day.

codelucas:
Good job, I'll port these changes to the python-2 branch today as well.

I was a bit concerned at first because I thought changing the HTML output that much in `Parser.nodeToString` would mess with the fulltext algorithm but apparently out 140 tests it didn't.

`nodeToString` is mostly used to get HTML output anyways (article HTML and video embed code).
I don't think we'd ever see a use case where `method` isn't equal to `'html'`, so good job!

ecesena:
Great to know that it didn’t take too much effort! I was likewise surprised to discover that <iframe/> is not valid html.

> On Feb 6, 2015, at 4:18 AM, Lucas Ou-Yang <notifications@github.com> wrote:
> 
> Good job, I'll port these changes to the python-2 branch today as well.
> 
> I was a bit concerned at first because I thought changing the HTML output that much in Parser.nodeToString would mess with the fulltext algorithm but apparently out 140 tests it didn't.
> 
> nodeToString is mostly used to get HTML output anyways (article HTML and video embed code).
> I don't think we'd ever see a use case where method isn't equal to 'html', so good job!
> 
> —
> Reply to this email directly or view it on GitHub.
> 

-- 
Emanuele Cesena <emanuele.cesena@gmail.com>
http://www.theneeds.com

Il corpo non ha ideali

PR #113, opened 2015-01-18
Improved title extraction via h1 and og:title

ecesena:
This commit improves title extraction by relying also on h1 and og:title. 
The current extraction was flawed in multiple ways: wrong part extracted during splitting, truncated title or typos or wrong spellings (probably caused by a bad CMS used by the news source).

In short, combining title, og:title, and h1 allows reducing these errors. The general assumption is that h1 is the most polished/curated title, as it's visible to the users, but can be hard to extract (not always present, not sure if we're detecting the right h1). og:title often contains the title properly split, and is also a good "3rd party" to help choosing among title and h1.

Explicit rules are documented within the get_title() comments.

These rules have been created via empiric tests (over 60 sources, about 10 articles/source). It's easy to find/create counter examples, but in general the results are better than before :) (about 100 more correct results over 4-500 articles). I tried to find a minimum set of understandable/reasonable rules that could cover the majority of the tests I had.

Included in the patch, I'm removing the split by ':' because most of the time this removes relevant part of the title (e.g., 'Game of the Week: Random Name', always removes the most significant part, i.e. the game name, and similar for recipes, events...). This potentially opens issues with titles like 'Report: This is the Real Title' or 'Watch: This is the Real Title'. However most of the time the use of og:title and h1 allows extracting only 'This is the Real Title' -- and when this doesn't happen, the source is using the complete string as title also for the users (h1), so in some sense we can say that the full string is the intended output.

As a future development, the split function could be improved by excluding the part(s) containing the source name (e.g., via the domain name). However, still from empirical tests, these are now rare cases (but existing, e.g., hgtv.com). IMO this can be ignored (too rare cases for too much extra code).

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -164,6 +164,20 @@ class ContentExtractor(object):
 
     def get_title(self, doc):
         """Fetch the article title and analyze it
+
+        Assumptions:
+        - title tag is the most reliable (inherited from Goose)
+        - h1, if properly detected, is the best (visible to users)
+        - og:title and h1 can help improve the title extraction
+        - python == is too strict, often we need to compare fitlered
+          versions, i.e. lowercase and ignoring special chars
+
+        Explicit rules:
+        1. title == h1, no need to split
+        2. h1 similar to og:title, use h1
+        3. title contains h1, title contains og:title, len(h1) > len(og:title), use h1
+        4. title starts with og:title, use og:title
+        5. title starts with h1 and h1 long enough, use h1
         """
         title = ''
         title_element = self.parser.getElementsByTag(doc, tag='title')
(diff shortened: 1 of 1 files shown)


yprez:
@codelucas , any chance you can review this?

ecesena:
Hi Yuri, great to hear that you're interested too :) As I was mentioning it's kind of hard to get good tests for it, and it's a really time consuming process. If you have some spare time and interest, I could provide (a lot of) data to set up tests. Let me/us know.

yprez:
@ecesena , Sure, I can take a look...
How is it currently tested?
Do you run it manually over a bunch of downloaded articles? Maybe it can just be automated somehow?

ecesena:
I extended newspaper's tests, but won't commit that because it would add tons of extra files. I have about 1k different sources with 2-10 articles/source. The process is kind of auto, but the issue is that it's not really clear what the correct result should be, unless you manually go through about 5k articles :) I did that for about 50-100 sources, then I stopped, refactored the code, submitted this pull req. Happy to share the test env privately, maybe drop me an email if needed (ec.at.theneeds.dot.com).

yprez:
@ecesena Sorry, I kinda forgot about this :)
Do you still use the same email? Or alternatively, can you send me those to y.at.yprez.com?

yprez:
Not 5000 articles, but I manually went through the first 100+ of the current fulltext tests (adding a script for title extraction only).

* @ecesena's version is generally better than the old parsing
* H1 usually contains the best result, but sometimes catches other unrelated titles from the text.
* og:title is usually a better candidate than title for the final parsing, but for some reason, sometimes not html-escaped well (probably because it's less visible then <title>). Maybe og:title is a better fallback when everything else fails.

I'll test a bit more, maybe add some titles that I know were wrongly parsed before, and see if they improve.

newspaper/extractors.py
```diff
@@ -175,48 +189,116 @@ def get_title(self, doc):
...
+        #     title_text = self.split_title(title_text, COLON_SPLITTER, title_text_h1)
+        #     used_delimeter = True
 
         title = MOTLEY_REPLACEMENT.replaceAll(title_text)
+
+        # in some cases the final title is quite similar to title_text_h1
+        # (either it differs for case, for special chars, or it's truncated)
+        # in these cases, we prefer the title_text_h1
+        filter_title = filter_regex.sub('', title).lower()
+        if filter_title_text_h1 == filter_title or filter_title_text_h1.startswith(filter_title):
```
yprez: If the `startswith` condition here is removed, it'll fix the error when unrelated stuff is appended to the titles coming from h1.

E.g. http://www.theglobeandmail.com/news/world/airasia-search/article22224253/
gets the title "Final manoeuvre of AirAsia flight raises questions over flying in storms Add to ..."

newspaper/extractors.py
```diff
@@ -164,6 +164,20 @@ def parse_byline(search_str):
...
+        - og:title and h1 can help improve the title extraction
+        - python == is too strict, often we need to compare fitlered
+          versions, i.e. lowercase and ignoring special chars
+
+        Explicit rules:
+        1. title == h1, no need to split
+        2. h1 similar to og:title, use h1
+        3. title contains h1, title contains og:title, len(h1) > len(og:title), use h1
+        4. title starts with og:title, use og:title
+        5. title starts with h1 and h1 long enough, use h1
```
yprez: I think that this rule (5) can cause lots of wrong results and a fallback to title is preferable here.

codelucas:
Thanks for your work on this @ecesena! @yprez's latest effort in #193 combines his and your effort so I'm closing this

PR #125, opened 2015-02-26
bumping nltk to 2.0.5 - see #824 in nltk

hexelon:
This has to be done due to https://github.com/nltk/nltk/issues/824

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ Pillow==2.5.1
 PyYAML==3.11
 cssselect==0.9.1
 lxml==3.3.5
-nltk==2.0.4
+nltk==2.0.5
 requests==2.3.0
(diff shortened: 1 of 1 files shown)


jwarzech:
+1 please pull this in as soon as you get a chance :)

codelucas:
Thanks for the PR guys! :+1: We should probably bump up the py3 nltk requirement to 3.0.1 as well ..

jwarzech:
Thanks!  When do you think you will do a release with this change?

codelucas:
A release should be out (for both py2 and py3) tonight! I'll update this thread when it's done.

codelucas:
New versions of newspaper are pushed live into pypi! Your pip installations should be working now.

Thanks so much for catching this early

PR #129, opened 2015-03-14
Adding setting for num of sentences in summary

mehuleo:
Sometime if Article is too long, 5 line summary will not do justice! So adding an extra setting in Configuration using this, one can set number of summary sentences.

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -320,7 +320,9 @@ class Article(object):
         keyws = list(set(title_keyws + text_keyws))
         self.set_keywords(keyws)
 
-        summary_sents = nlp.summarize(title=self.title, text=self.text)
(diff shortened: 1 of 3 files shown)


codelucas:
This looks great, will look at it/test before committing

newspaper/article.py
```diff
@@ -170,7 +173,11 @@ def parse(self):
         output_formatter = OutputFormatter(self.config)
 
         title = self.extractor.get_title(self.clean_doc)
-        self.set_title(title)
+        """if title has successfully extracted, 
+        override any existing title
+        """
+        if title is not None:
```
yprez: I think `if title:` would be better, in case it's an empty string

yprez: Actually, this isn't needed at all, because set_title() already checks this.

PR #135, opened 2015-04-03
Add python3-pip install step for Ubuntu

irnc:
Without `sudo apt-get install python3-pip` as an additional step, `pip3` command would not be available.

```
$ pip3 install newspaper3k
The program 'pip3' is currently not installed. You can install it by typing:
sudo apt-get install python3-pip
```


codelucas:
Good catch! Thanks for the PR

PR #144, opened 2015-05-13
Handle lxml raising ValueError on node.itertext()

### LINKED ISSUE #143: ValueError: Input object has no element: HtmlProcessingInstruction
For certain articles the exception `ValueError: Input object has no element: HtmlProcessingInstruction` is raised.

For example, for:

    from newspaper import Article
    article = Article(url='http://www.bostonglobe.com/news/world/2015/05/13/obama-discusses-iran-with-saudi-leaders/YQVleMtSURxsYgxUOznzFO/story.html')
    article.download()
    article.parse()

The traceback:

    /.../local/lib/python2.7/site-packages/newspaper/article.pyc in parse(self)
        221 
        222             text, article_html = output_formatter.get_formatted(
    --> 223                 self.top_node)
        224             self.set_article_html(article_html)
        225             self.set_text(text)

    /.../local/lib/python2.7/site-packages/newspaper/outputformatters.pyc in get_formatted(self, top_node)
         54         self.remove_trailing_media_div()
         55 
    ---> 56         text = self.convert_to_text()
         57         # print(self.parser.nodeToString(self.get_top_node()))
         58         return (text, html)

    /.../local/lib/python2.7/site-packages/newspaper/outputformatters.pyc in convert_to_text(self)
         61         txts = []
         62         for node in list(self.get_top_node()):
    ---> 63             txt = self.parser.getText(node)
         64             if txt:
         65                 txt = HTMLParser().unescape(txt)

    /.../local/lib/python2.7/site-packages/newspaper/parsers.pyc in getText(cls, node)
        194     @classmethod
        195     def getText(cls, node):
    --> 196         txts = [i for i in node.itertext()]
        197         return text.innerTrim(u' '.join(txts).strip())
        198 

    /.../local/lib/python2.7/site-packages/lxml/etree.so in lxml.etree._Element.itertext (src/lxml/lxml.etree.c:51098)()

    /.../local/lib/python2.7/site-packages/lxml/etree.so in lxml.etree.ElementTextIterator.__cinit__ (src/lxml/lxml.etree.c:68059)()

    /.../local/lib/python2.7/site-packages/lxml/etree.so in lxml.etree.iterwalk.__init__ (src/lxml/lxml.etree.c:131973)()

    /.../local/lib/python2.7/site-packages/lxml/etree.so in lxml.etree._rootNodeOrRaise (src/lxml/lxml.etree.c:14641)()

    ValueError: Input object has no element: HtmlProcessingInstruction

yprez:
Fixes #143

### DIFF
--- a/newspaper/outputformatters.py
+++ b/newspaper/outputformatters.py
@@ -60,7 +60,11 @@ class OutputFormatter(object):
     def convert_to_text(self):
         txts = []
         for node in list(self.get_top_node()):
-            txt = self.parser.getText(node)
+            try:
+                txt = self.parser.getText(node)
+            except ValueError:  # lxml error
+                txt = None
+
             if txt:
                 txt = HTMLParser().unescape(txt)
                 txt_lis = innerTrim(txt).split(r'\n')
--- a/newspaper/outputformatters.py
+++ b/newspaper/outputformatters.py
@@ -8,8 +8,12 @@ __license__ = 'MIT'
 __copyright__ = 'Copyright 2014, Lucas Ou-Yang'
 
 from HTMLParser import HTMLParser
+import logging
 
 from .text import innerTrim
+
+
+log = logging.getLogger(__name__)
 
 
 class OutputFormatter(object):
@@ -63,6 +67,7 @@ class OutputFormatter(object):
             try:
                 txt = self.parser.getText(node)
             except ValueError:  # lxml error
+                log.warning('Error parsing lxml node', exc_info=True)
                 txt = None
 
             if txt:


yprez:
Parsing now works:
```python
from newspaper import Article
article = Article(url='http://www.bostonglobe.com/news/world/2015/05/13/obama-discusses-iran-with-saudi-leaders/YQVleMtSURxsYgxUOznzFO/story.html')
article.download()
article.parse()

article.text
#[Out]# u'US President Barack Obama shook hands with Saudi Crown Prince Mohammed
#[bin Nayef as Foreign Minister Adel al-Jubeir (L) looked on
#[Wednesday.\n\nWASHINGTON (AP) \u2014 President Barack Obama welcomed Saudi
#[Arabian leaders to the White House for discussions on his overtures to
#[Iran.\n\nObama met in the Oval Office with the Saudi crown prince and deputy
#[crown prince on Wednesday.\n\nContinue reading below\n\nObama says the U.S.
#[and the Saudis are building on a strong relationship \u2018\u2018during a very
#[challenging time.\u2019\u2019\n\nThe president says they will discuss the
#[situation in Yemen as well as broader issues for the gulf region.\n\nThe Saudi
#[officials are representing the kingdom in Washington after King Salman
#[announced that he no longer would make the trip.\n\nLater Wednesday, leaders
#[from other Gulf nations with join Obama and the Saudis for a dinner at the
#[White House. They\u2019ll head to Camp David Thursday for broader talks on
#[Iran and regional security.'
```

newspaper/outputformatters.py
```diff
@@ -60,7 +60,11 @@ def get_formatted(self, top_node):
     def convert_to_text(self):
         txts = []
         for node in list(self.get_top_node()):
-            txt = self.parser.getText(node)
+            try:
+                txt = self.parser.getText(node)
+            except ValueError:  # lxml error
```
codelucas: I think we should at least print out warnings + fail loudly if this happens. Maybe notify the url that failed so the developer can triage further? The ideal is to debug urls that don't work so the library can slowly extract more successfully.

codelucas:
See my inline comment, otherwise thanks for this!

yprez:
@codelucas printing a warning makes sense here, but failing to parse the entire article because of one bad xml node is a bit drastic IMO.

Would you accept this fix if I add a `print()` or perhaps use `warnings.warn()` when this happens?
Any other way to push this forward?

codelucas:
@yprez Good point. Hmm, how about:

```
import traceback
traceback.print_exc()
```

A traceback will be a big/loud and also imformative warning that something is deeply wrong but won't kill the entire program.

Ideally we'd also output the URL that it failed on so the developer knows what went wrong and where to debug

yprez:
@codelucas Great. How about using `logging`, instead of printing to stdout?

Then we can use something like: `log.warning('error msg', exc_info=True)`, which will log the traceback as well.
What do you think?

yprez:
Also, I can't find a way to reach the `article` instance from `OutputFormatter`, so I can't append the URL when needed. And moving it to a higher level won't do the trick, because then the other nodes will be skipped, not just the broken one.

I'll make an intermediate commit with what I've got for now...

codelucas:
Thanks!

Issue #147, opened 2015-05-30
Author not found when element of class 'author' has children.


edouardklein:
Hi !

When the author's name is given in a child of an element with the appropriate class, get_authors() doesn't find it.

e.g. : 
`<div class="author"><span class="first">Albert</span><span class="last">Einstein</span></div>`

Using lxml's etree.tostring() function to get the html instead of the text attribute (which in this case is empty) could solve the problem, but I'm unsure whether it would create problems elsewhere. I'll propose a PR in this direction. For the specific case I listed above, it works, as parse_byline() removes the HTML markup anyway.

edouardklein:
I meant etree.tostring(), not etree() in my previous comment.

PR #148, opened 2015-05-30
Use children of suspected author element

edouardklein:
This may maybe solve issue #147

I'm sorry if it breaks anything else, I tried to make it keep the old behaviour : the ''.join(...) will give '' if there are no children, and it is not evaluated if the .text attribute is not empty.

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -20,6 +20,8 @@ import urllib.parse
 
 from tldextract import tldextract
 
+from lxml import etree
+
 from . import urls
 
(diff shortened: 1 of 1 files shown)


edouardklein:
There seem to be a problem calling str() on the return value of tostring().

For me, using etree.tostring(c).decode('utf-8') works, but one should use the actual encoding of the document, and I don't know how to access it.

codelucas:
This is a fairly sensitive change. Add unit tests to prove it solves issue #147? Otherwise I'm going to close for now

PR #149, opened 2015-05-31
bugfix in article.py, is_valid_body

ms8r:
Hi  - small fix in article.py, `is_valid_body` (compared `wordcount` rather than `len(wordcount)` against `MIN_WORD_COUNT`)


codelucas:
Wow, good find -- looks good to me!

PR #152, opened 2015-07-11
allow custom content_extractore

itech001:
the default content_extractor doesn't work for me, i need have one custom content extractor to work for my special task


newspaper/__init__.py
```diff
@@ -12,6 +12,9 @@
                   popular_urls, NewsPool, Configuration as Config)
 from .source import Source
 from .version import __version__
+from .extractors import ContentExtractor
+from .parsers import Parser
+from .network import get_html
```
yprez: Why add imports here?

yprez:
Why not set it outside of the library instead?
e.g.

```
article = Article()
article.extractor = CustomExtractor()

article.download()
#...
```

PR #170, opened 2015-11-17
Updated requests version

adrienthiery:
Proposing an update because requests-2.3.0 is creating an error for me on ubuntu 15.10
( like this one : https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=770157 )
I tried updating pyssl, urllib3 and such but only updating requests is working...


codelucas:
Thanks

PR #173, opened 2015-12-29
Add language: traditional chinese

evan176:


codelucas:
Languages codes are two character, can you find the right one for traditional Chinese? (I think the codebase restricts language inputs/selections to be two character codes).

We already have a `zh` stopwords file, if you find that's different than traditional Chinese rename that one.

codelucas:
Thanks for this

evan176:
Language codes indeed restricts it to be two character codes. I have no idea about proper name :(

yprez:
Hmm, it seems that it's "zh" for all "Chinese", both traditional and simplified. Kinda like "en-us" and "en-gb"...

http://stackoverflow.com/questions/4892372/language-codes-for-simplified-chinese-and-traditional-chinese

codelucas:
This isn't of the highest priority so I'm closing. Newspaper is in a phase where we want to just focus on fixing as many bugs and improving core scraping.

PR #176, opened 2016-01-05
Support urls with dots

alexanderlukanin13:
There is a critical bug in `newspaper3k==0.1.5` which prevents it from recognizing urls with dots in them.

For example: http://www.enfieldindependent.co.uk/news/

Please upload a new version on PyPI if possible.

### DIFF
--- a/newspaper/urls.py
+++ b/newspaper/urls.py
@@ -245,8 +245,13 @@ def url_to_filetype(abs_url):
         path = path[:-1]
     path_chunks = [x for x in path.split('/') if len(x) > 0]
     last_chunk = path_chunks[-1].split('.')  # last chunk == file usually
-    file_type = last_chunk[-1] if len(last_chunk) >= 2 else None
-    return file_type or None
+    if len(last_chunk) < 2:
+        return None
+    file_type = last_chunk[-1]
+    # Assume that file extension is maximum 5 characters long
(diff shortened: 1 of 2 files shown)


codelucas:
@alexanderlukanin13 another great find!

newspaper/urls.py
```diff
@@ -245,8 +245,13 @@ def url_to_filetype(abs_url):
         path = path[:-1]
     path_chunks = [x for x in path.split('/') if len(x) > 0]
     last_chunk = path_chunks[-1].split('.')  # last chunk == file usually
-    file_type = last_chunk[-1] if len(last_chunk) >= 2 else None
-    return file_type or None
+    if len(last_chunk) < 2:
+        return None
+    file_type = last_chunk[-1]
+    # Assume that file extension is maximum 5 characters long
+    if len(file_type) <= 5 or file_type.lower() in ALLOWED_TYPES:
```
codelucas: Perhaps 5 should be defined in a constants file?

alexanderlukanin13: @codelucas `5` is a quick-and-dirty heuristic, and it should be kept local. Moving it to constants would only give it an undeserved air of authority.

The proper solution is to use a extension blacklist and HEAD requests for uncertain cases.

PR #178, opened 2016-01-13
Handle lxml raising ValueError on node.itertext() - Python 3

yprez:
Same as #144 but on the Python3 branch.


codelucas:
Looks good to me!

Issue #174, opened 2015-12-30
Downgrades to old versions of required packages upon installation


cartisan:
The versions of the packages in requirements.txt are fixed with `==` .
This leads to the replacement of current versions of this packages with old ones upon installation of newspaper with `pip install newspaper`.

I would plead for using `>=` syntax for all required packages unless there is an important reason to use a specific version on single requirements. An nice example requirements.txt can be found here: https://github.com/scikit-image/scikit-image/blob/master/requirements.txt

codelucas:
Good idea, doing this now. (Originally done with == to ensure no breakage on current setups, but that clearly isn't working with many recent issues being around version numbers).

yprez:
+1
I can add this quickly (current versions and up should work).

codelucas:
Go for it @yprez

yprez:
@codelucas ping

PR #179, opened 2016-01-14
requirements.txt - Use minimal instead of exact versions

yprez:
requirements.txt - Use minimal instead of exact versions

As suggested in #174

I took the current version in requirements.txt (assuming nothing breaks with them) and changed them to the minimal version.
Downgrading to the real minimal version that works at this point is too complicated and might introduce new undiscovered bugs.

Tested locally on a fresh virtualenv.


codelucas:
This was so needed! Merging and uploading into pypi asap

PR #184, opened 2016-01-21
Fix installation docs

yprez:
Change according to README:

* Note about Python2 and 3.
* Change package name in pip install example to newspaper3k.


codelucas:
Thanks for this

PR #186, opened 2016-01-23
Added nltk named entities feature

makalaaneesh:
Along with ```article.keywords```,
we can now get all the named entities of the article by typing ```article.named_entities```
I thought this might be useful !


codelucas:
Since this is a larger change I'll have to investigate the quality of `named_entities` generation before accepting.

PR #187, opened 2016-01-25
Fix installation and build

yprez:
* Upgrade feedfinder2 version to fix installation (due to dfm/feedfinder2#5).
* Drop support for Python 3.2 (end of life in February 2016 and causing errors with some packages).
* Test on Python 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
- - "3.2"
  - "3.3"
  - "3.4"
 install:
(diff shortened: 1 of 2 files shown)


codelucas:
Good find on feedfinder, I'll have to update this in pypi tonight

yprez:
It's not such a huge issue now, it was broken between 0.0.2 and the next one, but should install ok now as is.

PR #188, opened 2016-01-26
Added Hebrew stop words for language support

alon7:
Hebrew support :)


yprez:
@alon7 I think that print_available_languages() in utils.py also needs to be updated: https://github.com/codelucas/newspaper/blob/master/newspaper/utils.py#L328

Other than that looks great to me.

alon7:
@yprez  Yeah, forgot to commit that file
Fixed it

thelonecabbage:
Amazing!!

PR #190, opened 2016-01-27
Don't scrape error pages

### LINKED ISSUE #142: Handling 404, 500, and other non-200 http response codes to prevent scraping error pages
Has it ever been considered that some links would respond with a non-200 response, and this module would end up scraping the contents of that error page? I'm looking at network.get_html() and if I have time, I might just PR a change to that method to accept a strict 200-only config. I'm posting this here just in case I'm missing something or if someone has handled this elegantly already. Thanks.

yprez:
Added a setting (default=True) to only scrape success pages. HTTP error pages are treated the same as other requests errors.

This fixes #142 

Also added minor code improvements (more strict exception catching and remove redundant `pass` statement).

### DIFF
--- a/docs/user_guide/advanced.rst
+++ b/docs/user_guide/advanced.rst
@@ -195,6 +195,8 @@ Here is a full list of the configuration options:
 
 ``keep_article_html``, default False, "set to True if you want to preserve html of body text"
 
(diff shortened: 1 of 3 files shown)


codelucas:
Looks good to me! Thanks for changing `except Exception` to `except requests.exceptions.RequestException`, catching base exceptions is bad form

PR #193, opened 2016-01-31
Improve title parsing

yprez:
This is @ecesena's work from PR #113 which I tweaked it a bit and tested.

Significant changes to #113 (the rest of the info is there): 

* Unescape html attributes. This fixes html entities appearing in og tags (og:title specifically).
* Title parsing - don't rely on h1 if long then title and og:title (often the added part is junk).

**Testing:**

I made a hacky title dump script to run over the fulltext articles urls and compared the results manually. On this subset of articles, this version gives the best results (although not perfect).

The results in an ugly spreadsheet here: https://docs.google.com/spreadsheets/d/1eUakaWAdldyjFd-cp-OaqmI-vUCgooXQfLZcmcsATa8/edit?usp=sharing

Contains the follwoing sheets:

1. "Fulltext titles" the same articles as in fulltext tests - went over these manually.
2. "short titles" - articles/pages from my projects database, that had short urls. these are usually parsed wrong so I compared the 2 versions of the script on those. There are over 500 results with different titles so I didn't compare them manually, but in overall, it looks like the new version is better.

Color codes:

* Green - good result
* Yellow - not the best but acceptable
* Red - wrong.

(Adding extra columns would be more efficient, but this was the quickest way I could think of)

@codelucas if you could review that would be great.

If anyone has any ideas of how to improve and/or test this, let me know :)

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -237,6 +237,7 @@ class ContentExtractor(object):
         2. h1 similar to og:title, use h1
         3. title contains h1, title contains og:title, len(h1) > len(og:title), use h1
         4. title starts with og:title, use og:title
+        5. use title, after splitting
         """
         title = ''
         title_element = self.parser.getElementsByTag(doc, tag='title')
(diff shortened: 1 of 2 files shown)


yprez:
Also, I have a database with over 10 mil extracted articles and other metadata (without the original html, but they could be parsed again). These could be used for testing titles... just need to think of a sane subset of these and maybe a better way to verify what consists of a good title.

codelucas:
Thanks for following up with this PR, I'll take a closer look tonight!

yprez:
By the way, I've been using this in production for 3 weeks now, and the results are better than before.

yprez:
@codelucas, I rebased this against master so it applies cleanly again.

codelucas:
Wow nicely done, the google doc you made was a great illustration of why this is better than what we currently have in master

ecesena:
Hey, just saw this. Congrats for the merging and thanks so much @yprez for moving forward with this!

yprez:
Woop! awesome!

Thanks @codelucas, @ecesena !

I'll try improving title parsing further now that this one is merged...

PR #194, opened 2016-02-01
 Adding setting for num of sentences in summary

yprez:
Updated version of PR #129 by @mehuleo, plus minor changes and docs.

> Sometime if Article is too long, 5 line summary will not do justice! So adding an extra setting in Configuration using this, one can set number of summary sentences.

The original branch isn't there anymore so I applied the commits with `git am`. Looks like it worked fine.


codelucas:
LGTM, thanks for this catch

PR #197, opened 2016-02-06
Ignore punctuation when calculating link density

krzd:
When calculating link density, punctuation characters should not count as words.
This could be if a paragraph consists of a single link followed by punctuation such as a full stop.
The punctation could also consist of multiple characters such as an ellipsis.
In that case it should still be regarding as a having a high density of links.

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -863,18 +863,20 @@ class ContentExtractor(object):
         link to text ratio, then the text is less likely to be relevant
         """
         links = self.parser.getElementsByTag(e, tag='a')
-        if links is None or len(links) == 0:
+        if not links:
             return False
 
         text = self.parser.getText(e)
-        words = text.split(' ')
+        words = filter(lambda word: word.isalnum(), text.split())
+        if not words:
+            return True
         words_number = float(len(words))
         sb = []
         for link in links:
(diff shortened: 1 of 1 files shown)


codelucas:
Thanks for this PR! And great catch. However I have a few questions:

How will this affect foreign languages? The method you are using inspects if the word is alphanumeric chars only via `isalnum`. `u'字'.isalnum()` evaluates to True, which is good but we should have unit tests.

Can you add unit tests for an example with foreign languages and also add an example where ignoring punctuation generates a correct summary (while including punctuation yields a bad summary?)


Thanks a lot again for this, great catch!

krzd:
@codelucas 
This should work for all languages in the same way.
As per documentation (https://docs.python.org/3/library/stdtypes.html#str.isalnum) the lookup is done by checking the character's Unicode class. Any kind of digit or letter, defined as such in Unicode, would therefore be regarded as alpha-numeric by `isalnum()`.
The code performing the lookup can also be inspected here: https://github.com/python/cpython/blob/1fe0fd9feb6a4472a9a1b186502eb9c0b2366326/Objects/unicodectype.c#L291.

I will try to dig out the example that caused me to create this patch and add unit tests soon.

codelucas:
Thanks for this!

Issue #200, opened 2016-02-10
Improve test assertions


yprez:
Currently, when the some assertion in the tests are failing, the reason isn't shown and it requires further debugging.

For example:

```
Traceback (most recent call last):
  File "tests/unit_tests.py", line 136, in runTest
    self.test_nlp_body()
  File "tests/unit_tests.py", line 43, in run
    method(*args, **kw)
  File "tests/unit_tests.py", line 262, in test_nlp_body
    assert self.article.summary == SUMMARY

AssertionError
```

This is because the `assert` statement is used with the regular `unittest`, instead of of `assertEquals`, `assertTrue` or other unittest methods.

I see two solutions to this.

1. Use py.test with the current tests.
2. Switch to using unittest assertEquals statements (ugly and outdated imo).

The main feature of py.test is dealing with assert statements as test assertions. If failing, it'll show the exact drilldown of why the assertions failed, down to the single missing character (for example).

It shouldn't be too hard to switch to py.test from the current test runner, and I think it'll improve the test suite a lot, and will make it easier to write new tests when needed. I noticed that tests are often skipped for new features, maybe this can help...

yprez:
@codelucas I still would prefer py.test, but if you're happy with the changes in #203 this can be closed.

yprez:
After #203 `ExhaustiveFullTextCase` is executed every time.
Maybe it should be split into another file, and executed from there when needed.

PR #203, opened 2016-02-12
Unittests refactoring

amorgun:
I've made tests independent and used unittest test loader.
I also have changed assert statements as suggested in #200.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -98,6 +98,7 @@ class ExhaustiveFullTextCase(unittest.TestCase):
         except Exception:
             print('<< URL: %s parse ERROR >>' % url)
             traceback.print_exc()
+            pubdate_failed, fulltext_failed = True, True
         else:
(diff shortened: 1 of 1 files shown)


codelucas:
Thanks for refactoring this, looks good to me

tests/unit_tests.py
```diff
@@ -477,51 +496,34 @@ def test_chinese_fulltext_extract(self):
...
+        self.assertEqual(text, fulltext(article.html, 'es'))
 
 
 if __name__ == '__main__':
-    # unittest.main()  # run all units and their cases
-
-    suite = unittest.TestSuite()
-
-    if len(sys.argv) > 1 and sys.argv[1] == 'fulltext':
-        suite.addTest(ExhaustiveFullTextCase())
```
yprez: @amorgun @codelucas Removing this makes the tests runs "fulltext" every time... not sure if this is the best approach...
In any case, the readme should be updated to reflect this.

Issue #175, opened 2016-01-01
get_canonical_link should check og:url meta attribute


AlJohri:
From the Open Graph protocol:

> og:url - The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "http://www.imdb.com/title/tt0117500/".

I'd be happy to submit a PR if you comment below. Thanks!

yprez:
+1

codelucas:
Go ahead!

PR #204, opened 2016-02-13
get_canonical_link() method - return og:url if available

yprez:
As proposed in #175 

If `rel=canonical` is not available, try to get the open graph equivalent: `og:url`.


codelucas:
LGTM, thanks for also refactoring

PR #207, opened 2016-02-18
Fixed logging messages and level + comment fix.

AndreCAndersen:
In the function set_reddit_top_img of article.py I edited the logging messages such that they made more sense, as well as not giving it such a high logging level. We were getting critical errors about "jpeg error with PIL, timed out". However, it seemed like it was simply images.py which was having a time out issue when downloading images for dimension measuring. I don't believe it had anything to do with PIL.

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -371,10 +371,12 @@ class Article(object):
         except TypeError as e:
             if "Can't convert 'NoneType' object to str implicitly" in e.args[0]:
                 log.debug("No pictures found. Top image not set, %s" % e)
+            elif "timed out" in e.args[0]:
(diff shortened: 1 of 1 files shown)


codelucas:
Perhaps you can add an extra `if` clause in the `except` to catch the case where it's a timeout or other unimportant errors and log it as a debug. But if a serious PIL error happens (which I recall they do if a user has issues installing) we do want to log as critical.

AndreCAndersen:
No problem! Love to help out. Made another commit based on your feedback. Any thoughts?

codelucas:
Looks good to me! Thanks

Issue #215, opened 2016-02-24
Suggestion: images.py - use requests instead of urllib for consistency


yprez:
And cleaner code.

yprez:
Also, chunk size isn't respected: https://github.com/codelucas/newspaper/blob/89848c976bd9edfa8e90a45fab24baa34af058a6/newspaper/images.py#L109

Issue #210, opened 2016-02-21
Parsing issue - BadStatus line when parsing articles


yprez:
Looks like it's caused by some images returning an invalid http status code...

The code:
```
import newspaper
article = newspaper.Article(url='http://www.himalayanwindowsinc.com/itineraries/show/253')
article.download()
article.parse()
```

The error:
```
Traceback (most recent call last):
  File "/home/yuri/virtualenvs/newspaper/lib/python3.5/site-packages/IPython/core/interactiveshell.py", line 3066, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-df704f976c28>", line 1, in <module>
    article.parse()
  File "/home/yuri/repos/newspaper/newspaper/article.py", line 228, in parse
    self.fetch_images()
  File "/home/yuri/repos/newspaper/newspaper/article.py", line 247, in fetch_images
    self.set_top_img(first_img)
  File "/home/yuri/repos/newspaper/newspaper/article.py", line 415, in set_top_img
    if s.satisfies_requirements(src_url):
  File "/home/yuri/repos/newspaper/newspaper/images.py", line 227, in satisfies_requirements
    img_url, self.useragent, referer=self.url)
  File "/home/yuri/repos/newspaper/newspaper/images.py", line 170, in fetch_image_dimension
    return fetch_url(url, useragent, referer, retries, dimension=True)
  File "/home/yuri/repos/newspaper/newspaper/images.py", line 102, in fetch_url
    open_req = opener.open(req, timeout=5)
  File "/usr/lib/python3.5/urllib/request.py", line 465, in open
    response = self._open(req, data)
  File "/usr/lib/python3.5/urllib/request.py", line 483, in _open
    '_open', req)
  File "/usr/lib/python3.5/urllib/request.py", line 443, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.5/urllib/request.py", line 1268, in http_open
    return self.do_open(http.client.HTTPConnection, req)
  File "/usr/lib/python3.5/urllib/request.py", line 1243, in do_open
    r = h.getresponse()
  File "/usr/lib/python3.5/http/client.py", line 1174, in getresponse
    response.begin()
  File "/usr/lib/python3.5/http/client.py", line 282, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.5/http/client.py", line 264, in _read_status
    raise BadStatusLine(line)
http.client.BadStatusLine: 
```

[...]

PR #216, opened 2016-02-25
images.py - replace urllib with requests

yprez:
Using `response.raw` to make it as close as possible to the previous behavior.

In the future the code can be refactored further...

### DIFF
--- a/newspaper/images.py
+++ b/newspaper/images.py
@@ -12,11 +12,9 @@ import logging
 import math
 import io
 import traceback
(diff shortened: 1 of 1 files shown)


yprez:
Closes #215

yprez:
Apparently this also fixes #210.

I came upon images.py when trying to investigate it, but I didn't think refactoring this would solve it...

codelucas:
Amazing work, you're killing it @yprez

PR #218, opened 2016-02-27
Tests - Fix skipping ExhaustiveFullTextCase

yprez:
Revert the previous behavior of skipping ExhaustiveFullTextCase, unless `fulltext` is passed as an argument.

Ref: #203, #200

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -2,7 +2,6 @@
 """
 All unit tests for the newspaper library should be contained in this file.
 """
-import logging
 import sys
 import os
 import unittest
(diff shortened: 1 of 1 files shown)


amorgun:
I think you shouldn't skip tests ever. `ExhaustiveFullTextCase` basically works fine and I can't see any reason why not to run it on every build.

yprez:
With `ExhaustiveFullTextCase` the tests run more than twice as long as without it (for me it's 20 seconds vs. 50 seconds). I guess that was the reason for excluding it.

I'm fine with whatever @codelucas decides.

If `ExhaustiveFullTextCase` is included in every run the readme should be updated instead.

codelucas:
I removed this test a few months ago -- when they were flaky. But now they work fine, thanks guys

PR #219, opened 2016-02-27
Improved title splitting and tests

### LINKED ISSUE #73: Closing quotation mark is removed from title
I work mostly with Russian articles. In Russian, «angled» quotes are main variation of quotation marks. So, I noticed that if there's «angled» quotes in a title of an article, all closing quotation marks are removed from extracted title and it contains only the opening ones. I found that in happens [here](https://github.com/codelucas/newspaper/blob/d6d16134d082f8d9e2b7b62c42265dbea73baa32/newspaper/extractors.py#L214) in `ContentExtractor`:

    TITLE_REPLACEMENTS = ReplaceSequence().create(u"&raquo;").append(u"»")
	
	...
	
	return TITLE_REPLACEMENTS.replaceAll(title).strip()
	
As far as I understand, it's needed for removing `»` from titles where this character is used as a delimiter. Maybe it'll make sense to modify the regular expressions to not remove right quotes that have left quotes before them?

[Here](http://vibirai.ru/articles/vyigray_bilety_na_oficialnuu_premeru_neuderjimye_3_v_kinomaks_ur-1201706)'s an example of a page that have broken quotation marks in extracted title (Russian language).

yprez:
* Add some tests for title extraction - testing the extractor directly, in order not to add many heavy "fulltext" tests.
* Fix `TITLE_REPLACEMENTS` behavior which was removing both "&raquo;" and "»" instead of replacing the first with the second.
* Fix #73 by surrounding ARROW_SPLITTER with spaces. Usually when used as a splitter it will be preceded and followed by a space.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -298,6 +298,26 @@ class ArticleTestCase(unittest.TestCase):
         self.assertCountEqual(KEYWORDS, self.article.keywords)
 
 
+class ContentExtractorTestCase(unittest.TestCase):
+    """Test specific element extraction cases"""
+
+    def setUp(self):
+        self.extractor = newspaper.extractors.ContentExtractor(Configuration())
+        self.parser = newspaper.parsers.Parser
+
+    def _get_title(self, html):
+        doc = self.parser.fromstring(html)
+        return self.extractor.get_title(doc)
+
+    def test_get_title_basic(self):
+        html = '<title>Test title</title>'
+        assert self._get_title(html) == 'Test title'
+
+    def test_get_title_split(self):
+        html = '<title>Test page » Test title</title>'
+        assert self._get_title(html) == 'Test title'
+
+
 class SourceTestCase(unittest.TestCase):
     @print_test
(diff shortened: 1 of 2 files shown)


tests/unit_tests.py
```diff
@@ -298,6 +298,35 @@ def test_nlp_body(self):
...
+        self.extractor = newspaper.extractors.ContentExtractor(Configuration())
+        self.parser = newspaper.parsers.Parser
+
+    def _get_title(self, html):
+        doc = self.parser.fromstring(html)
+        return self.extractor.get_title(doc)
+
+    def test_get_title_basic(self):
+        html = '<title>Test title</title>'
+        assert self._get_title(html) == 'Test title'
```
amorgun: Wrong kind of assertion, should be `self.assertEqual('Test title', self._get_title(html))`

yprez: Right. Too used to py.test I suppose...

codelucas:
Nicely done!

PR #221, opened 2016-03-06
Fix unicode encoding errors in get_parsing_candidate

### LINKED ISSUE #211: Parsing issue - UnicodeEncodeError in get_parsing_candidate()
```
import newspaper
article = newspaper.Article(url='http://waynemessick.com/marketing/small-business-internet-marketing-advertising-strategies')
article.download()
article.parse()
```

Causes:
```
Traceback (most recent call last):
  File "/home/yuri/repos/newspaper/newspaper/article.py", line 166, in parse
    parse_candidate = self.get_parse_candidate()
  File "/home/yuri/repos/newspaper/newspaper/article.py", line 333, in get_parse_candidate
    return RawHelper.get_parsing_candidate(self.url, self.html)
  File "/home/yuri/repos/newspaper/newspaper/utils.py", line 59, in get_parsing_candidate
    raw_html = raw_html.encode('utf-8')
UnicodeEncodeError: 'utf-8' codec can't encode character '\ude7e' in position 13277: surrogates not allowed
```

yprez:
The encoding here is done for generating a hash of the content so ignoring a few incorrect unicode characters is better than throwing exceptions.

Fixes #211.

### DIFF
--- a/newspaper/utils.py
+++ b/newspaper/utils.py
@@ -47,8 +47,8 @@ class FileHelper(object):
 
 class ParsingCandidate(object):
 
-    def __init__(self, urlString, link_hash):
-        self.urlString = self.url = urlString
+    def __init__(self, url, link_hash):
+        self.url = url
         self.link_hash = link_hash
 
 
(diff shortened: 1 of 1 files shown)


codelucas:
Good catch on `self.urlString` being unused!

PR #224, opened 2016-03-10
Update requirements to NLTK v3.2

alvations:
NLTK 3.1 and 3.2 has quite a lot of bug fixes. Hope you don't mind if the requirement for your library uses the newer NLTK version.


yprez:
Apparently nltk didn't officially support Python 3.5 before 3.1 (http://www.nltk.org/news.html). So I think it's worth to upgrade.

Thanks!

codelucas:
:+1:

PR #249, opened 2016-05-14
Added Hindi language stopwords

parulsethi:
Added Hindi language stopwords.
sources: http://www.ranks.nl/stopwords/hindi


agnelvishal:
It looks like tokenizer has not been added. Consider looking http://tdil-dc.in/index.php?option=com_download&task=showresourceDetails&toolid=1515&lang=en

agnelvishal:
Also tamil stop words can be added from http://tdil-dc.in/index.php?option=com_download&task=showresourceDetails&toolid=1641&lang=en

Issue #260, opened 2016-06-06
Redirect should follow meta refresh


adamn:
If newspaper goes to a page like this:

https://www.google.com/url?rct=j&sa=t&url=http://sfbay.craigslist.org/eby/cto/5617800926.html&ct=ga&cd=CAAYATIaYTc4ZTgzYjAwOTAwY2M4Yjpjb206ZW46VVM&usg=AFQjCNF7zAl6JPuEsV4PbEzBomJTUpX4Lg

It receives HTML like this:

```html
<script>window.googleJavaScriptRedirect=1</script><script>var n={navigateTo:function(b,a,d){if(b!=a&&b.google){if(b.google.r){b.google.r=0;b.location.href=d;a.location.replace("about:blank");}}else{a.location.replace(d);}}};n.navigateTo(window.parent,window,"http://sfbay.craigslist.org/eby/cto/5617800926.html");
</script><noscript><META http-equiv="refresh" content="0;URL='http://sfbay.craigslist.org/eby/cto/5617800926.html'"></noscript>
```
Which I got from a Google Alert feed:

https://www.google.com/alerts/feeds/02224275995138650773/15887173320590421756

Then it does not follow the meta refresh link inside the HTML.

The underlying Requests library can't see HTML so I think it makes sense for Newspaper to handle this situation with a new flag (follow_meta_refresh ?) that would default to False because of the performance implications.

yprez:
@adamn newspaper is using `requests` for getting the html, it doesn't execute any javascript on the page. It causes the redirect bugs and some issues with dynamic pages.

If you want to download dynamic pages, you can try downloading the html with a headless browser like phantom.js and then pass it's results to newspaper. Parsing JS is a big issue, but I don't see it being added to newspaper anytime soon.

yprez:
Maybe there's a way to plug a different method of downloading articles than `requests`, without breaking the current API. That could be useful...

adamn:
Sorry, I wasn't thinking when I wrote the ticket.  What I meant to imply is that this would have to be a condition attached to download(), not parse() - and so should default to False since afaik download() doesn't parse the html currently.

However, for users that want to follow the meta refresh tag, it could really make life easier if Newspaper handled it rather than each dev writing custom code.

I would not suggest interpreting any javascript, just parsing the html and looking for `<meta http-equiv="refresh" />` and parsing the content value.

[...]

PR #263, opened 2016-06-10
Meta refresh redirect support 

logandhead:
I made an optional argument in the confugration that supports redirects that are stated in meta_refresh tags in accordance to these specs https://www.w3.org/TR/WCAG20-TECHS/H76.html and this issue https://github.com/codelucas/newspaper/issues/260

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -180,6 +180,8 @@ class ArticleTestCase(unittest.TestCase):
 
     @print_test
     def test_meta_refresh_redirect(self):
+        # TODO: We actually hit example.com in this unit test ... which is bad
+        # Figure out how to mock an actual redirect
         config = Configuration()
         config.follow_meta_refresh = True
         article = Article(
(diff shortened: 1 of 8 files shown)


codelucas:
Thanks for the PR! This mostly looks good to me but I'll let @yprez give the final call on merging this

codelucas:
If he doesn't respond in a few days I'll merge

adamn:
I upped the version to 0.1.8 so we could install on our systems more easily.  I'm happy to switch that around or revert.

tests/unit_tests.py
```diff
@@ -179,6 +179,17 @@ def test_download_html(self):
...
     @print_test
+    def test_meta_refresh_redirect(self):
+        config = Configuration()
+        config.follow_meta_refresh = True
+        article = Article(
+            '', config=config)
+        html = mock_resource_with('google_meta_refresh', 'html')
+        article.download(html=html)
+        article.parse()
+        self.assertEqual(article.title, 'Example Domain')
```
yprez: Does this actually pull from example.com? I'm not sure it's such a good idea to do that during tests, but can't think of other alternatives that will fit into the current mocking pattern...

yprez: I guess we can stick a TODO comment here for now and merge this.

adamn: TODO is now there.

PR #269, opened 2016-07-08
Re-implemented parse_feeds() and removed Python 3.3

adamn:
- Removed Python 3.3 so I could use [html.unescape](https://docs.python.org/3/library/html.html#html.unescape)
- `source.parse_feeds()` now attaches `title` to feeds list.
- Did not use `extractor.get_title()` to get the title since it didn't work well with RSS feeds.

I'm going to need some help on how best to write tests.  It looks like there isn't much in the way of tests right now for feeds so thoughts are appreciated.


adamn:
Any input on how to go about tests on this?

newspaper/parsers.py
```diff
@@ -247,7 +246,7 @@ def getAttribute(cls, node, attr=None):
         if attr:
             attr = node.attrib.get(attr, None)
         if attr:
-            attr = HTMLParser().unescape(attr)
+            attr = unescape(attr)
```
codelucas: Good change - it seems like HTMLParser().unescape(..) is being deprecated in 3.5+ and we will need to use unescape instead

codelucas: I would have preferred the style of 
```
import html
html.unescape(..)
```

But it's not like this repository does a good job of setting standards :( I'm going to accept this and merge it for sake of removing the deprecating warnings hitting everyone and probably refactor later

newspaper/source.py
```diff
@@ -204,21 +232,22 @@ def parse_categories(self):
 
         self.categories = [c for c in self.categories if c.doc is not None]
 
+    def _map_title_to_feed(self, feed):
```
codelucas: This is a really interesting interpretation of feed parsing! I'll give it a pass for now and do some testing on it later to see how it performs.

newspaper/source.py
```diff
@@ -133,7 +129,39 @@ def set_feeds(self):
         """Don't need to cache getting feed urls, it's almost
         instant with xpath
         """
-        urls = self.extractor.get_feed_urls(self.url, self.categories)
+        common_feed_urls = ['/feed', '/feeds', '/rss']
```
codelucas: Great heuristics to catch extra feeds. Thanks for this. I'll probably test this new code and add unit tests for the new edge case catches later. Nice job

requirements.txt
```diff
@@ -1,12 +1,12 @@
 beautifulsoup4>=4.4.1
-Pillow>=2.6.1
+Pillow>=3.3.0
```
codelucas: Thank you for doing this!

newspaper/article.py
```diff
@@ -144,17 +144,21 @@ def build(self):
         self.parse()
         self.nlp()
 
-    def download(self, html=None, title=None):
+    def download(self, html=None, title=None, recursion_counter=0):
         """Downloads the link's HTML content, don't use if you are batch async
         downloading articles
+
+        recursion_counter (currently 1) stops refreshes that are potentially
```
codelucas: Is 1 the best value? I think this should be in the config file.

professorgobblefarts (commented):
#itsAboutTime

PR #271, opened 2016-07-12
Fix caption

vishaltak:
The original caption removal code does not remove a lot of image captions which ultimately become a part of the article text and mess it up. Have added a fix to do the same.


ArielFrischer:
Can someone consider updating this with latest newspaper3k if it is working well?

lethaiq:
Thanks @vishaltak  for the helpful code

lethaiq:
> The original caption removal code does not remove a lot of image captions which ultimately become a part of the article text and mess it up. Have added a fix to do the same.

thank you

Issue #161, opened 2015-09-09
How can I load html downloaded page?


abhigenie92:
Does this 
    article.download()
equalivent to
  article.html=html_downloaded

And later I can parse and get article text.

younes0:
a = Article('')
a.set_html(html)
a.parse()

This was already mentionned in: #74

yprez:
can be closed

Issue #74, opened 2014-08-14
Parsing Raw HTML


lhausmann:
Is it possible to send raw html directly to the Article.parse() function without it being downloaded by Article.download()?

Issue #37, opened 2014-02-05
How to assign html content without downloading it?


dizlv:
Is it possible to assing already downloaded html string to the Article object without calling download() method?

Because I want to use it in Scrapy project and html page is already downloaded so I need simply to parse it.

PR #275, opened 2016-07-16
Add `html` kwarg for loading article content

lwm:
My personal experience, #74, #161 and #37 bring me to believe this is necessary.

As it stands right now, I don't understand why you force the user to `download()`.

It's a confusing API (using `set_html`) for people who just want the parse functionality.

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -33,7 +33,7 @@ class ArticleException(Exception):
 class Article(object):
     """Article objects abstract an online news article page
     """
-    def __init__(self, url, title='', source_url='', config=None, **kwargs):
+    def __init__(self, url, html='', title='', source_url='', config=None, **kwargs):
         """The **kwargs argument may be filled with config values, which
         is added into the config object
         """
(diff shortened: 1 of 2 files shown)


lwm:
bump!

newspaper/article.py
```diff
@@ -89,14 +89,20 @@ def __init__(self, url, title='', source_url='', config=None, **kwargs):
...
         self.article_html = ''
 
         # Flags warning users in-case they forget to download() or parse()
         # or if they call methods out of order
         self.is_parsed = False
+
         self.is_downloaded = False
+        if self.html:
+            # User has loaded html content, does not require download restriction
+            self.is_downloaded = True
```
codelucas: Why did you add `self.is_downloaded = True`? set_html sets this to False as intended

codelucas:
I added inline comments on why you set the is_downloaded flag. Is there a big need for this API change? I don't want to change the API often after this has been out so long unless there is a real need

newspaper/article.py
lwm: If the user passes some value for `self.html`, then we want to pass the `is_downloaded` check.

lwm:
>  Is there a big need for this API change?

You had 3 issues and I also ran into the issue. More will come.

>  I don't want to change the API often after this has been out so long

This isn't a breaking change. It just allows people *not* using the download path.

PR #278, opened 2016-08-05
Added Japanese language support

Newspad:
I've added Japanese language support and it works fine.
This change doesn't affect any other code.


codelucas:
This looks great, I'm about to merge and accept but can you just add one unit test where 'ja' language works?

Newspad:
I've added a unit test for 'ja' language as you requested and replaced `html.parser.HTMLParser` with `html.parser` to avoid deprecated method warnings.
In addition, I've wrapped the test code in try-catch blocks because Travis can't run MeCab parser and rejects my code.

evilplanet:
Any progress on this front? Adding ja support would be great!

evilplanet:
@Newspad So Japanese text extraction is only supported on CentOS? If I don't use `.nlp()` method, can I use Japanese article extraction on heroku-powered projects?

newspaper/text.py
```diff
@@ -156,3 +156,65 @@ def get_stopword_count(self, content):
...
+                        mecab.parse('')
+                        node = mecab.parseToNode(stripped_input)
+                        while node:
+                            word = node.surface.lower()
+                            pos = node.feature.split(',')[0]
+                            if len(word) > 0 and not pos == '記号':
+                                words.append(word)
+                            node = node.next
+        except Exception as e:
+            pass
```
codelucas: instead of `pass` and we please add exception logging in a pretty format saying Japanese stopwords extraction has failed. (in your experience does `candidate_words` have a high rate of failure? If not I'd even suggest no exception catching here.

newspaper/text.py
```diff
@@ -156,3 +156,65 @@ def get_stopword_count(self, content):
...
+    """Japanese segmentation
+    """
+    def __init__(self, language='ja'):
+        super(StopWordsJapanese, self).__init__(language='ja')
+
+    def candidate_words(self, stripped_input):
+        words = []
+        try:
+            import MeCab
+            import os, subprocess
```
codelucas: Can we have a separate try/catch for these two import statements and print a pretty "you must install mecab .. etc" if it is not installed?

README.rst
```diff
@@ -265,6 +266,25 @@ NOTE: If you find problem installing ``libpng12-dev``, try installing ``libpng-d
     $ curl https://raw.githubusercontent.com/codelucas/newspaper/master/download_corpora.py | python3
 
 
+**If you are on CentOS** and use Japanese language support, install using the following:
```
codelucas: Not clear if this means:
1) Japanese language support works by default or all non-CentOS operating systems and you don't need to install
2) Japanese language support works ONLY on CentOS, and if you are on CentOS install.

If 2) is true, I think this design is not ideal. A language support should be for all operating systems, not just one because a parse/tokenizing library is available only on that OS.

vitaly-zdanevich:
For AWS Lambda anyway we will put all dependencies into deployment zip - so MeCab will be there too.

asafcombo:
Hi Any updates on JP support ?
If this is not going to happen soon, should I use [https://github.com/Newspad/newspaper](url) instead ?

evilplanet:
@asafcombo The problem with Japanese support is you have to install MeCab, which is not easy feat. It really depends on your environment. You could take Newspad's code and make sure MeCab is installed on your server. Change the code where MeCab is called if necessary. (I use natto-py to call MeCab)

codelucas:
https://github.com/codelucas/newspaper/pull/584 is merged! Giving us JP support. Thanks again for this initial effort here @Newspad but I decided to merge PR-584 as it has minimal dependency requirements. Also unfortunately we don't have a good way to measure branch vs. branch performance in article extraction quality (will file a todo for myself)

PR #228, opened 2016-03-24
Fix relative canonical link

yprez:
Fixes #227

PR #296, opened 2016-10-14
Fix relative canonical link (See #228)

adamn:
PR #228 looks good but it wasn't handling a meta url like 'example.com/article.html', only urls like 'article.html' would have been parsed correctly.

I added a new test and refactored `get_canonical_link()` accordingly

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -515,20 +515,18 @@ class ContentExtractor(object):
         result = ''
 
         links = self.parser.getElementsByTag(doc, tag='link', attr='rel', value='canonical')
-        canonical = self.parser.getAttribute(links[0], 'href') if links else ''
 
+        canonical = self.parser.getAttribute(links[0], 'href') if links else ''
         og_url = self.get_meta_content(doc, 'meta[property="og:url"]')
 
-        if canonical:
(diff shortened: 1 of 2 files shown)


yprez:
Looks good to me.
@codelucas can you take a look as well?

codelucas:
Great catch! Also thanks so much for also adding unit tests 👍 
Merged.

newspaper/extractors.py
```diff
@@ -511,26 +511,35 @@ def get_canonical_link(self, article_url, doc):
...
+                            format(parsed_article_url.hostname),
+                                  parsed_meta_url.path)
+                try:
+                    true_path = strip_hostname_in_meta_path.group(1)
+                except AttributeError:
+                    true_path = parsed_meta_url.path
+
+                # true_path may contain querystrings and fragments
+                meta_url = urlunparse((parsed_article_url.scheme,
+                                       parsed_article_url.hostname,true_path,
```
codelucas: I would have put true_path on its own line, but thanks for fixing my python style in this 👍

adamn: Yeah that was me just using cmd+shift+l in PyCharm.

PR #301, opened 2016-11-11
fix DATE REGEX

sbagan:
While processing urls like this http://www.sltrib.com/sltrib/mobile3/58201224-219/laws-animal-farm-gag.html.csp current date regular expression catches "201224" which is just part of some internal id, so article['publish_date'] becomes 2024-12-20.


codelucas:
Great catch on this!
I will accept this immediately after a unit test is added, feel free to use the example you gave in the comment above with "http://www.sltrib.com/sltrib/mobile3/58201224-219/laws-animal-farm-gag.html.csp" not working. See tests.py

codelucas:
I can help resolve merge conflicts ^

PR #302, opened 2016-11-12
Make python3k instructions more prominent in docs

IllegalCactus:
Mention the instructions for installing on Python 3.x next to the Python 2.x command, so that the difference between `newspaper` and `newspaper3k` is less easy to miss.

Also removes some trailing whitespace.

### DIFF
--- a/docs/user_guide/install.rst
+++ b/docs/user_guide/install.rst
@@ -23,15 +23,19 @@ However, you will run into fixable issues if you are trying to install on ubuntu
 
 - For PIL to recognize .jpg images::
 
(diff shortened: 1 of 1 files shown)


codelucas:
Rejecting for now

IllegalCactus:
No worries, makes sense :)

yprez:
@codelucas I noticed that many people open issues, because they installed the wrong version of newspaper on Python 3. Maybe just put the current py2/py3 explanation higher up in the README?
Or maybe move the the "getting newspaper" section above the examples?

codelucas:
Let's try a doc-fix for now, this is an important call, potential customers will be turned off if an install fails at the very start

PR #305, opened 2016-11-22
use http as default protocol

hslee16:
otherwise the source_url cannot be set


codelucas:
Good catch, not sure why this edge case was not noticed / caught earlier? Are schemas being `None` common? Can you log this and check

codelucas:
It would be helpful if you can add unit tests for this specific scenario also @hslee16

codelucas:
Feel free to open an issue for that

Issue #291, opened 2016-10-03
extract content from a html file


crosofg:
How to extract content from html file generated by headless browsers like pyqt

yprez:
Try `set_html()`, Something like:

```
article = newspaper.Article()
article.set_html(html)

article.parse()
article.nlp()
```

crosofg:
Now I am getting other error
```
Traceback (most recent call last):
  File "/home/ubuntu/workspace/newspaperrepo/newspaperExample.py", line 1, in <module>
    from newspaper import Article
  File "/usr/local/lib/python2.7/dist-packages/newspaper3k-0.1.8-py2.7.egg/newspaper/__init__.py", line 10, in <module>
    from .article import Article, ArticleException
  File "/usr/local/lib/python2.7/dist-packages/newspaper3k-0.1.8-py2.7.egg/newspaper/article.py", line 12, in <module>
    from . import images
  File "/usr/local/lib/python2.7/dist-packages/newspaper3k-0.1.8-py2.7.egg/newspaper/images.py", line 15, in <module>
    import urllib.parse
ImportError: No module named parse
```

yprez:
Try `Article(url='')` or `Article(url='http://example.com/test-url')`

yprez:
@crosofg this should do it, feel free to reopen if it still doesn't work.

km5ar:
this method above doesn't seems to be working anymore, any thoughts? 

article = newspaper.Article(url = "")
article.set_html(html)
article.parse()
article.text

PR #307, opened 2016-11-26
Added explaination on using exisiting html string

### LINKED ISSUE #300: Parse from a string instead of URL
I see no interface to parse from a user-supplied string of HTML. Because of #299 we can't use the regular downloader so I thought we could load text ourselves and pass the HTML to an Article but it doesn't seem to be possible?

Bartvds:
Code from #291

Closes #300

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -104,6 +104,22 @@ A Glance:
     u'The study shows that 93% of people ...'
 
 
+Newspaper can directly parse a string:
+
+.. code-block:: pycon
+
+    >>> from newspaper import Article
+
+    >>> a = Article('http://example.com/fake')  # Pass any url
+
+    >>> my_html = u'<html><title> ...'
+
+    >>> a.set_html(my_html)  
+
+    >>> a.parse()    
+    >>> ...
+
+
 Newspaper has *seamless* language extraction and detection.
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -104,6 +104,22 @@ A Glance:
     u'The study shows that 93% of people ...'
 
 
+Newspaper can directly parse a string:
+
+.. code-block:: pycon
+
+    >>> from newspaper import Article
+
+    >>> a = Article('http://example.com/fake')  # Pass any url
```
dustyny: I'm not sure I would be able to understand this without having context. Can you expand upon it just a little? 
> Pass any url, valid or not. This will be overridden in the next steps.

docs/index.rst
```diff
@@ -104,6 +104,22 @@ A Glance:
...
+
+    >>> my_html = u'<html><title> ...'
+
+    >>> a.set_html(my_html)
```
dustyny: I think it would be helpful to comment this line as well. 

> This is where you will pass in your file to override the URL you passed in the previous step.

codelucas:
Thanks for the effort on this issue @Bartvds and @dustyny!

I will accept+merge after the issues from @dustyny are addressed

PR #308, opened 2016-12-01
added stopwords for polish language

deer667:
I would like to use your module in my project and would like to have better handling of polish language. List of words based on http://www.ranks.nl/stopwords/polish, https://github.com/bieli/stopwords and https://pl.wikipedia.org/wiki/Wikipedia:Stopwords


codelucas:
Fantastic, thank you so much for this. Can you also add a link to add "pl" in the list of available languages like done here?: https://github.com/codelucas/newspaper/commit/3cd0e2c486ebd6f17d9489738d83bf1c81106282

Please copy the format of above ^, that way the library is more helpful to users, so they know `pl` is a valid language!

codelucas:
I will accept after this is done

PR #313, opened 2016-12-28
Changes in Readme file

valaparthvi:
I was facing problem while install "libpng12-dev", so I tried installing "libpng-dev" instead, which worked fine. So I added a little note in the Readme file.
 
screenshot from 2016-12-28 20-29-42

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -241,6 +241,8 @@ package name is ``newspaper``.
 
     $ sudo apt-get install libjpeg-dev zlib1g-dev libpng12-dev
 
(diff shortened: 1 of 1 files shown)


codelucas:
Cool stuff, merging. Thanks for the pr. Would you mind also checking to see why libpng12-dev failed and libpng-dev, if we can narrow it down to an OS and also add that into the readme it would be ideal!

valaparthvi:
Happy to :) I'll update the details soon.

PR #324, opened 2017-01-25
Meta image logic fix

mercuree:
1. Fixed condition:
```python
if try_one is None:
```
which is always False, because 
```python
 try_one = self.get_meta_content(doc, 'meta[property="og:image"]')
```
never returns None

2. Changed the condition order. Icon image is least relevant for top_img

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -431,7 +431,7 @@ class ContentExtractor(object):
         try_one = self.get_meta_content(doc, 'meta[property="og:image"]')
         if try_one == '':
             link_img_src_kwargs = \
-                {'tag': 'link', 'attr': 'rel', 'value': 'img_src'}
+                {'tag': 'link', 'attr': 'rel', 'value': 'img_src|image_src'}
             elems = self.parser.getElementsByTag(doc, **link_img_src_kwargs)
             try_two = elems[0].get('href') if elems else None
 
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -430,18 +430,18 @@ class ContentExtractor(object):
         top_meta_image, try_one, try_two, try_three, try_four = [None] * 5
         try_one = self.get_meta_content(doc, 'meta[property="og:image"]')
         if try_one == '':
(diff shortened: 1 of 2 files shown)


newspaper/extractors.py
```diff
@@ -429,19 +429,19 @@ def get_meta_img_url(self, article_url, doc):
         """
         top_meta_image, try_one, try_two, try_three, try_four = [None] * 5
         try_one = self.get_meta_content(doc, 'meta[property="og:image"]')
-        if try_one is None:
-            link_icon_kwargs = {'tag': 'link', 'attr': 'rel', 'value': 'icon'}
-            elems = self.parser.getElementsByTag(doc, **link_icon_kwargs)
-            try_two = elems[0].get('href') if elems else None
-
-        if try_two is None:
+        if try_one == '':
```
yprez: `if not try_one` would be more elegant imo

yprez: Would also make the next conditions simpler.

mercuree: You are right. When you replace `if try_one is none` with `if not try_one`, the difference is unclear, and i wrote more explicit way.

codelucas:
This is a fantastic fix @mercuree. Also great work on the unit tests 👍 💯 Double checking and merging it now!

codelucas:
Wait, this appears to be failing the CI build - the failure is in the new unit test you added: 
https://travis-ci.org/codelucas/newspaper/jobs/279784081

Weird how it passed the CI build locally in the comment though

codelucas:
I'll take a look

codelucas:
Haha problem found! Ironically the new bug was due to an optimization from your earlier PR!

Your PR here (https://github.com/codelucas/newspaper/pull/355) improved our xml parse optimization by turning regex off .. BUT this PR was based on the code before your optimization and thus does not turn on the regex api.

new code in this PR:
```
 elems = self.parser.getElementsByTag(doc, **link_img_src_kwargs)
```

I'll submit a PR later tonight before deploying a new version that re-enables regex operations (or performs two queries)

PR #329, opened 2017-02-05
Previous siblings were inserted in reverse order

mercuree:
`add_siblings()` adds previous node siblings (which were also scored as good elements) before top_node.
Usually, such nodes don't exist or there is only one node, which is previous sibling of top_node and has good score. So this code worked well in many cases. But if we have more than one such node, the bug appears because node elements are inserted always at first position in a list `top_node.insert(0, p)`, so first element becomes last and vise versa.
It was hard to catch this bug, 70 web pages from different domains just passed all my tests with this fix and without it.
I will add tests later,


mercuree:
Added tests

louyangsc:
Thank you for creating this PR @mercuree - I'm investigating the change now!

yprez (approved):
Looks good to me

PR #339, opened 2017-03-01
Guess character encodings using HTML tag "meta".

alosultan:


newspaper/parsers.py
```diff
@@ -48,7 +48,17 @@ def get_unicode_html(cls, html):
...
         if not html:
             return html
-        converted = UnicodeDammit(html, is_html=True)
+
+        # Guess encoding using "meta" tag.
+        guessed_encodings = None
+
+        document = lxml.html.fromstring(str(html))
+        meta_tag = document.xpath("//meta[@charset] | //meta[@http-equiv='Content-Type']")
+        if meta_tag:
```
mercuree: I think, this part is redundant. UnicodeDammit searches meta charset internally by regular expression, when is_html flag is set. You don't need to parse full document.

PR #354, opened 2017-03-28
Use Temp Dir instead of Home Dir

dvf:
Using home directories is bad practice for certain deployment strategies (Elastic Beanstalk, Heroku etc), and limits the OS-scope of the project. Rather use a Temp Directory, which is more secure and doesn't require extra permissions (for a server role eg. Elastic Beanstalk)

### DIFF
--- a/newspaper/settings.py
+++ b/newspaper/settings.py
@@ -10,6 +10,7 @@ __copyright__ = 'Copyright 2014, Lucas Ou-Yang'
 
 import logging
 import os
+import tempfile
 
 from http.cookiejar import CookieJar as cj
 
@@ -31,7 +32,7 @@ NLP_STOPWORDS_EN = os.path.join(
 
 DATA_DIRECTORY = '.newspaper_scraper'
 
-TOP_DIRECTORY = os.path.join(os.path.expanduser("~"), DATA_DIRECTORY)
+TOP_DIRECTORY = os.path.join(tempfile.gettempdir(), DATA_DIRECTORY)
 if not os.path.exists(TOP_DIRECTORY):
     os.mkdir(TOP_DIRECTORY)
 


dvf:
@codelucas I wasn't sure if there were contribution guidelines. 😕

s-alty:
@dvf Thanks for this. This solved an issue we were having when we tried to use newspaper on AWS Lambda.

dvf:
@s-alty yeah, ran into same issue. :)

stenbln:
If you deploy it on AWS put the following in container commands:



```
  00_write_permission_for_temp_newspaper_folder:
    command: "chmod o+w /tmp/.newspaper_scraper"
    ignoreErrors: true
```

This adds the needed write permission for the `/tmp/.newspaper_scraper` folder

dvf:
*bump* @codelucas

codelucas:
This is a great change, thanks @dvf 💯 👍 ! Do you guys see any downsides of this api?

One problem (assuming that the tempdir api call changes the temp directory every time) is that newspaper will lose its longterm article/news source caching ability. But i'm not sure if the API changes the temp path, will have to check the python source

dvf:
Thanks @codelucas 👍 

`/tmp` is at the mercy of the OS. Since we're using a named directory, my guess is that it's not more likely to be deleted than a file anywhere else.

dvf:
@codelucas *bump*

jnyborg:
I would also love to see this change.

yprez:
Would like to see a pull request for this. Sounds pretty easy to change.

I personally don't have the time to work on this (and I assume @codelucas also doesn't :) )

jnyborg:
The gettempdir does not change the temp dir across multiple calls. Looking at the [docs](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir) and the [source](https://github.com/python/cpython/blob/3.6/Lib/tempfile.py#L163), it just returns whatever the different TMP environment variables points to, so as long as these remain the same, it should work for long term caching, in the case where the user does not reboot or otherwise clean the temporary files. In that case, it might be fine anyways as the user probably wanted a fresh state.

yprez:
Why not default to `home` and use another env var if the user wants it to point to another dir? any downsides?

jnyborg:
That would be perfect. Perhaps something like this? 

```python
TOP_DIRECTORY = os.environ.get("NEWSPAPER_DATA_DIR") or os.path.join(
    os.path.expanduser("~"), DATA_DIRECTORY)
```

yprez:
I would just assing ~ to /tmp, but I'm lazy...

dvf:
What about support for Windows?

codelucas:
bitmoji

codelucas:
Will push a new pypi version for this tonight (along with some changes to allow users to override)

svemaraju:
So was this never pushed to PyPi? 🤔

PR #355, opened 2017-03-29
Using simple xpath inside getElementsByTag

mercuree:
Using xpath "contains" is much faster than regular expression.
My tests are running about 23s instead of 32s before this fix (30% faster)
However, we can still use regular expression as an option by passing an argument `use_regex=True`


codelucas:
Thanks for this @mercuree - will take a look.

codelucas:
This looks great, thanks! 💯 Hmm, you mention that using xpath contains improves the speed by 30% without side-effects? In that case what's the point if even including the use_regex=True? Are there any advantages to regex?

newspaper/parsers.py
```diff
@@ -108,13 +109,18 @@ def getElementById(cls, node, idd):
...
         # selector = tag or '*'
         selector = 'descendant-or-self::%s' % (tag or '*')
         if attr and value:
-            selector = '%s[re:test(@%s, "%s", "i")]' % (selector, attr, value)
-        elems = node.xpath(selector, namespaces={"re": NS})
+            if use_regex:
+                NS = {"re": "http://exslt.org/regular-expressions"}
+                selector = '%s[re:test(@%s, "%s", "i")]' % (selector, attr, value)
+            else:
+                trans = 'translate(@%s, "%s", "%s")' % (attr, string.ascii_uppercase, string.ascii_lowercase)
```
codelucas: you are translating uppercase to lowercase before the contains check. Does that fully imitate the effects of the regex `re:test`?

mercuree: This immitates "i" flag in regular expression (case insensitive), which was used before.

mercuree:
> In that case what's the point if even including the use_regex=True? Are there any advantages to regex?

I checked all usages of this method in entire project (at the moment i wrote this patch) and did not find any regular expression cases (just simple strings), but sometimes it may be useful, and i use regexp in my fork
https://github.com/mercuree/newspaper/blob/b03430c1847be7c5c9a5dc3721d62ea12e9e87b1/newspaper/extractors.py#L456

codelucas:
looks good to me! merging now, thanks for this work 👍 💯

PR #356, opened 2017-03-30
NLTK Corpus dependency install before newspaper3k

12DReflections:
The corpora is required for the install or newspaper3k or else it errors.
I tried pip installing, cloning & downloading older versions of newspaper3k before realising.

While the change seems minor it may prevent users from being deterred because they couldn't get the library working, thought it would be a good idea :-)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -243,14 +243,13 @@ package name is ``newspaper``.
 
 NOTE: If you find problem installing ``libpng12-dev``, try installing ``libpng-dev``.
 
-- Install the distribution via pip::
-
-    $ pip3 install newspaper3k
-
 - Download NLP related corpora::
 
     $ curl https://raw.githubusercontent.com/codelucas/newspaper/master/download_corpora.py | python3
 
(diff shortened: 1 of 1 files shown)


thundergolfer:
Do you mean that `pip3 install newspaper3k` fails for you if you don't first download the corpora?

I just tried to run that command and it installed OK. Or do you mean that the library is non-functional before you download the corpora, even though it may have installed successfully?

12DReflections:
That's correct, on Mac OS X El Capitan `pip3 install newspaper3k` had a dependency issue and wouldn't successfully install on the system without the corpora. 
I can retry at the office in a virtual environment.

codelucas:
The corpera should not block the installation, rather it blocks functionality inside newspaper. 

Can you please paste the precise installation errors you get in shell here so we can see the sequence of events? Thanks!

I'm going to merge this anyways since it's probably more expected for our users to have all functionality of newspaper working (which does require the corpera)

Good catch, thanks again

PR #358, opened 2017-04-06
Trailing slash for url date regex

akionakamura:
Take a look at this post: http://www.mildicasdemae.com.br/2017/04/5-coisas-que-irritam-o-seu-bebe-e-como-acabar-com-o-choro.html

The current code detects the date from the URL as `2017-04-05`, but the `5` is actually part of the title. I'm assuming that after the date patterns in the URL path it is expected a slash. With the rising of list posts, I think this might be relevant.


codelucas:
This is a great heuristic, thanks for adding it @akionakamura! 💯 👍 ! Let me take a look at the unit tests before accepting and merging.

codelucas:
I would accept this a lot quicker but the DATE_REGEX isn't just used for publishing date extraction. It is actually one of the main heuristics in our core url news identification algorithm here:
https://github.com/codelucas/newspaper/blob/ee20d92a03f840ebc8e5ecc0899c8addac72fe37/newspaper/urls.py#L196-L227

I'm scared this will affect article identification performance.
Debating merging this anyways and creating a separate regex for the article identification code

akionakamura:
Cool, thanks for the feedback.

PR #373, opened 2017-05-29
Added Greek language support

tseste:
I would like to use your module in a project that needs to support Greek language.
I have checked it in Greek webpages and it works fine.


tseste:
Hey @codelucas, could you check this one please, if there is any available time of course.
Thank you in advance.

codelucas:
Thanks so much for doing this @tseste! merging now (and also deploying a new version to pip tonight with your changes)

I'm kinda surprised that the Greek stopwords file is so large, 1,000+ stopwords is a lot bigger than the standard English stopword sets!

tseste:
Thank you so much @codelucas.

kickapoo:
@tseste 
Can you please add also this:
[stopwords-el-complex.txt](https://github.com/codelucas/newspaper/files/1049657/stopwords-el-complex.txt)

There are some without word accent and some more complex using Katharevousa greek system (https://en.wikipedia.org/wiki/Katharevousa)

FYI : @codelucas

Issue #357, opened 2017-04-02
You must `download()` an article before calling `parse()` on it!


hoommaah:
i have a problem with parsing articles and i think its because i placed parse right after downloading the article. do you think there is a chance that the article is not yet done downloading when i started parsing it? any suggestions? thanks!

yprez:
Newspaper is downloading articles synchronously, so that's probably not the issue.
Can you post the code that you're running?

yprez:
Newspaper is downloading articles synchronously, so that's probably not the issue.
Can you post the code that you're running?

craymichael:
I can reproduce this error consistently; it seems as if an Exception should be raised, or maybe even `None` returned instead (i.e. page not recognized as an article or valid page)?

Minimum steps to reproduce (Python 3):
```>>> from newspaper import Article
>>> a = Article('')
>>> a.download()
>>> a.parse()
You must `download()` an article before calling `parse()` on it!
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/newspaper/article.py", line 168, in parse
    raise ArticleException()
newspaper.article.ArticleException

>>> a = Article('http://www.irishtimes.com/business/sectors/financial-services/ecb-to-reveal-bad-loan-hurdles-for-euro-zone-bank-test-1.1719128')
>>> a.download()
>>> a.parse()
You must `download()` an article before calling `parse()` on it!
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/newspaper/article.py", line 168, in parse
    raise ArticleException()
newspaper.article.ArticleException
```

toshiro92:
Same issue for me, I am on Python3.5, and even I called `download()` before `parse()`, it shows me the same error.

Helw150:
@craymichael The Irish Times link you placed there is 404ing, which will intentionally cause this error.

However, I do have several instances of non-404ing links that raise this error. I believe it has something to do with pages with slow load times.

codelucas:
I'll take a look later, thanks for reporting this guys

codelucas:
This is being addressed right now in https://github.com/codelucas/newspaper/pull/378

codelucas:
Hey all, happy to announce that this has been FIXED in the latest version of newspaper3k in pip (version 2.0.1) pull now!

```
FAIL EXAMPLE: 

[...]

PR #378, opened 2017-06-10
Add logic to handle http != 2XX response failures. Inform user download happened

codelucas:
This is a much needed feature due to a lot of issues being created where people are calling `download()` on their article objects but fail to `parse()` the article contents from the HTML due to bad API design not distinguishing good http responses and bad http responses.

This PR makes it clear when an http response to some news website fails - whether by rate limiting, broken website server, etc. Now our api users will be more clear of what is going on.

Waiting for current integration/unit tests to pass and thinking about adding more

cc https://github.com/codelucas/newspaper/issues/357

### DIFF
--- a/newspaper/network.py
+++ b/newspaper/network.py
@@ -18,6 +18,8 @@ from .settings import cj
 log = logging.getLogger(__name__)
 
 
+FAIL_ENCODING = 'ISO-8859-1'
+
 def get_request_kwargs(timeout, useragent):
     """This Wrapper method exists b/c some values in req_kwargs dict
     are methods which need to be called every time we make a request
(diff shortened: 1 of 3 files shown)


codelucas:
Unit tests (which cover the html fetching path) are passing. I'm going to leave this up for another few hours before merging and updating pip

newspaper/network.py
```diff
@@ -31,44 +33,52 @@ def get_request_kwargs(timeout, useragent):
...
-            return response.text
-        return response.content
+        return _get_html_from_response(response)
 
     try:
-        html = None
-
         response = requests.get(
             url=url, **get_request_kwargs(timeout, useragent))
+    except requests.exceptions.RequestException as e:
```
codelucas: This is important. STILL catch `requests` exceptions in this function, but DON'T CATCH the 4XX or 5XX response codes. This disambiguation will allow our callers to have the HTTPException when a 4XX or 5XX happens and act accordingly

PR #383, opened 2017-06-14
Fix gravity score conversion

### LINKED ISSUE #382: remove_negativescores_nodes fails with invalid literal for int
```
  File "/usr/lib/python3.6/site-packages/newspaper/article.py", line 235, in parse
    self.top_node)
  File "/usr/lib/python3.6/site-packages/newspaper/outputformatters.py", line 48, in get_formatted
    self.remove_negativescores_nodes()
  File "/usr/lib/python3.6/site-packages/newspaper/outputformatters.py", line 109, in remove_negativescores_nodes
    score = int(score, 0)
ValueError: invalid literal for int() with base 0: '97.0'
```

Will submit a PR shortly that should fix it.

jamesmallen:
Fixes #382

### DIFF
--- a/newspaper/outputformatters.py
+++ b/newspaper/outputformatters.py
@@ -106,7 +106,7 @@ class OutputFormatter(object):
             self.top_node, "*[gravityScore]")
         for item in gravity_items:
             score = self.parser.getAttribute(item, 'gravityScore')
-            score = int(score, 0)
+            score = score and float(score) or 0
             if score < 1:
                 item.getparent().remove(item)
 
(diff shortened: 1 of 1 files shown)


codelucas:
This is a great catch thanks @jamesmallen, my original use of `and` instead of ternary was boneheaded. Did you notice this bug happening on a significant # of articles? Or very rare one-offs?

This logic seems to pass the unit tests - so I'm going to accept and merge! Push to pypi.

jamesmallen:
I only noticed it on a very small # of articles, and it was after I was monkeying around with using the html5lib parser instead of the default lxml one (which might turn into another PR). I'm not sure why it wasn't hitting this before, though!

codelucas:
This might need to be revisited in the future. I wonder if defaulting to 0 is ideal or seeing why gravity is causing problems and giving `None`s

PR #385, opened 2017-06-15
Add JSON-LD support by using extruct

torbenbrodt:
You will realize that publish_date for a some major German brands is empty

```python
url = "www.spiegel.de/politik/deutschland/terror-und-innere-sicherheit-die-neue-deutsche-haerte-a-1152113.html"
article = Article(url)
article.download()
article.parse()

article.title
>>> Steinmeiers Bundespräsidialamt: Personalrat tritt geschlossen zurück

article.publish_date
>>> None
```

The publisher is using JSON-LD, so I wondered how newspaper could support this. Actually there is a library called https://github.com/scrapinghub/extruct which has similar dependencies but comes with a jsonld extractor.

The MR is a proof of concept. Tests are missing, etc.
Happy to get your thoughts.


requirements.txt
```diff
@@ -10,3 +10,4 @@ tldextract>=2.0.1
 feedfinder2>=0.0.4
 jieba3k>=0.35.1
 python-dateutil>=2.5.3
+extruct
```
codelucas: Don't just list the package name, list >=SPECIFIC_VERSION to finely define what we want. i.e. find out the latest version of extruct that you're using right now and set this to be >= that version

newspaper/extractors.py
```diff
@@ -21,6 +21,11 @@
 from tldextract import tldextract
 from urllib.parse import urljoin, urlparse, urlunparse
 
+from extruct.jsonld import JsonLdExtractor
```
codelucas: Can we get away with this by just using the libraries newspaper already imports (e.g. lxml, BeautifulSoup) I'm not against adding extruct but it would be better to keep this library with fewer dependencies

newspaper/extractors.py
```diff
@@ -195,6 +199,16 @@ def parse_date_str(date_str):
             if datetime_obj:
                 return datetime_obj
 
+        JSONLD_TAGS = [
+            'datePublished'
+        ]
+        jsonlde = JsonLdExtractor()
+        jsonlddata = jsonlde.extract_items(doc)
```
codelucas: Is this call expensive? I worry since we are running this over the entire html `doc`

codelucas (commented):
Wow @torbenbrodt this is great work - thanks for doing this.

Please see my inline comments + add some preliminary tests for this functionality. (If you aren't familiar with how to run tests online feel free to use our travis CI build system)

newspaper/extractors.py
torbenbrodt: I don't like dependencies too. But I found the implementation quite difficult.

newspaper/extractors.py
torbenbrodt: I expect yes.

But I think the json-ld is widely used nowadays. If we decide to integrate json-ld into newspaper, we will probably need to restructure the code and move some priority of these parsers up.

torbenbrodt:
Thx for the fast feedback. What do you think? Tests are green? Let's include extruct?

mccarran:
Is this enhancement still being considered? I am finding many major publishers using ld+json for publish dates and much more, which comes up missing with current implementation. Thank you.

codelucas:
@mccarran @torbenbrodt This is still being considered, but I'm considering various implementations, will accept or make comments later

torbenbrodt:
Hey @codelucas , any update on this?
Actually extruct did make good progress recently and now includes further formats so I really suggest including it

agnelvishal:
```
#!/usr/bin/python

import newspaper
from newspaper import Article
from bs4 import BeautifulSoup
import json

url="https://yourstory.com/2018/11/leadership-john-chambers-cisco/"
article = Article(url)
article.download()
article.parse()

parsedHTML = BeautifulSoup(article.html,"lxml")
scripts = parsedHTML.find_all('script', type='application/ld+json')
for script in scripts:
    try:
    	data = json.loads(script.text)
    	print(data['author']['name'])
    	print(data['datePublished'])
    except:
        pass
```
The above code can also can be used. However I am having trouble integrating with newspaper3k.

simonm3:
Would be great if this could be completed. Many major media sites such as the bbc use json_ld and it would enable newspaper to extract publish_date.

Kerl1310 (approved):
LGTM

Kerl1310:
Fix the conflicting requirements.txt and I'd say this is good to go

PR #392, opened 2017-07-01
Fixing typo on line 82

jboo1212:


codelucas:
Thanks for the fix @jboo1212! ;)

PR #398, opened 2017-07-11
Extended user guide for adding new language support

tramwaj29:


codelucas:
Thanks for doing this @tramwaj29!

PR #417, opened 2017-08-07
Update requirements.txt

andreyfedoseev:
Use minimal instead of exact versions


codelucas:
Thanks for doing this @andreyfedoseev! 💯 however a change adding this has already been merged a few days ago

PR #418, opened 2017-08-17
auto decode requests response

lossme:


codelucas:
Thanks for your PR! I'm not familiar with this api - will need to study and get back to you on this

codelucas:
Can you please give an example of how this change will help, where otherwise it would fail?

lossme:
Response object has a apparent_encoding function, apparent_encoding by calling chardet.detect () to identify the text encoding, but it should be noted that this is some consumption of computing resources.

```   
    @property
    def apparent_encoding(self):
            """The apparent encoding, provided by the chardet library."""
            return chardet.detect(self.content)['encoding']


    @property
    def text(self):
        content = None
        encoding = self.encoding

        if not self.content:
            return str('')

        if self.encoding is None:
            encoding = self.apparent_encoding

        try:
            content = str(self.content, encoding, errors='replace')
        except (LookupError, TypeError):
            content = str(self.content, errors='replace')

        return content

```
After reading the source code，this work has been done by r.text
It seems that what I do is superfluous.

codelucas:
Great point, after looking at Requests' documentation is does appear that `.text` API call also does what you do.

http://docs.python-requests.org/en/master/api/#requests.Response.text

Thanks for filing regardless!

PR #419, opened 2017-08-28
Added 'Hindi' language

aeroaks:
Added hindi language by using basic stopwords text file. Did not find a
natural processing based stemmer. example used is:
'http://www.bhaskar.com'


codelucas:
Thanks for doing this @aeroaks. Will take a look and run tests to confirm it works! :) 👍 💯

newspaper/resources/text/stopwords-hi.txt
```diff
@@ -0,0 +1,43 @@
+को
```
codelucas: I am going to LGTM this and merge it in, my only comment is that the hindi stopwords list found on other sites seems larger: https://github.com/stopwords-iso/stopwords-hi/blob/master/stopwords-hi.txt

How did you determine that this is the best performing list?

newspaper/text.py
```diff
@@ -156,3 +156,28 @@ def get_stopword_count(self, content):
         ws.set_stopword_count(len(overlapping_stopwords))
         ws.set_stop_words(overlapping_stopwords)
         return ws
+
+
+class StopWordsHindi(StopWords):
```
codelucas: I just realized that `StopWordsKorean` and your new `StopWordsHindi` dont' need separate classes, both classes reuse 100% of the code from `StopWords` - it's fine I will merge this PR but assign myself a task to dedupe the code later

newspaper/text.py
```diff
@@ -156,3 +156,28 @@ def get_stopword_count(self, content):
...
+    """Hindi segmentation
+    """
+    def __init__(self, language='hi'):
+        super(StopWordsHindi, self).__init__(language='hi')
+
+    def get_stopword_count(self, content):
```
codelucas: Also I'm not sure if Hindi is a language that simply tokenizes words by spaces (like English) or needs more complex tokenizing like Arabic/Chinese.

See: https://github.com/taranjeet/hindi-tokenizer. Do you have any advice? (I don't know Hindi at all)

codelucas:
Thanks for this great work! @aeroaks

newspaper/resources/text/stopwords-hi.txt
aeroaks: This seems to be a more comprehensible list!

newspaper/text.py
aeroaks: I am new to NLP. Will have a look into this library!

aeroaks:
Thanks for accepting it. I am new to how NLP works. If you have already created an Issue then I can start with that or I will make an issue with the changes and updates that you have suggested.

PR #421, opened 2017-08-30
Fix article URL generation

fengkaijia:
Homepage URL as ```source_url``` in article creation will cause ```prepare_url``` to render a wrong URL in some scenarios. For example:

* Homepage: ```http://example.com/```
* Category: ```http://example.com/cat/```
* ```a href``` in the category page: ```./123/```
* Real Article URL (expected): ```http://example.com/cat/123/```
* URL detected: ```http://example.com/123/``` (by combining ```http://example.com/``` and ```./123/```)

### DIFF
--- a/newspaper/source.py
+++ b/newspaper/source.py
@@ -261,7 +261,7 @@ class Source(object):
             for url in urls:
                 article = Article(
                     url=url,
-                    source_url=self.url,
+                    source_url=feed.url,
                     config=self.config)
                 cur_articles.append(article)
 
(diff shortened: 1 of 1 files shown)


codelucas:
Thanks for doing this @fengkaijia, will take a look / run tests and get back to this PR soon 💯

codelucas:
WOW this is a terrible bug, great fix! I am merging immediately and will probably deploy a new version just for this..

We should also add unit tests for this very case of 

Homepage: http://example.com/
Category: http://example.com/cat/
a href in the category page: ./123/
Real Article URL (expected): http://example.com/cat/123/

Do you mind doing it? Or should I?

fengkaijia:
@codelucas I added #441, but it does not test if the source urls are extracted correctly. One way to do that is to change the ```test_source_build``` which use currently CNN (that doesn't use relative URL). To add a test case for this problem, I think we need to replace CNN with (or add) an another site.

PR #422, opened 2017-08-30
Speed up getElementsByTags

megatron-me-uk:
The `css_select` method internally converts the css selector string into an xpath selector string and then calls xpath. It is quite a lot faster to simply use an xpath selector. In profiling, this change reduces `getElementsByTags` from ~ 25% of `article.parse` execution time to less than 10%.
This PR should not change functionality.

### DIFF
--- a/newspaper/parsers.py
+++ b/newspaper/parsers.py
@@ -162,12 +162,9 @@ class Parser(object):
 
     @classmethod
     def getElementsByTags(cls, node, tags):
-        selector = ','.join(tags)
-        elems = cls.css_select(node, selector)
-        # remove the root node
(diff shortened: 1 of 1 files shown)


codelucas:
This is a great hack @megatron-me-uk! Thanks for this PR 👍 💯 ! The existing unit tests which test for parsing seem to have passed so I am accepting and merging into master.

This change should make it into `pip` in a few days, perhaps in the future a script showing the performance improvements will help the PR!

mercuree:
@codelucas 
Please, see also this pr: https://github.com/codelucas/newspaper/pull/355

PR #423, opened 2017-08-30
Do not remove a node if it contains an article

megatron-me-uk:
Sometimes the cleaner can be a little overzealous and delete a node that contains the main article node. This change may save the article in simple cases.

### DIFF
--- a/newspaper/cleaners.py
+++ b/newspaper/cleaners.py
@@ -46,6 +46,7 @@ class DocumentCleaner(object):
             .create("\n", "\n\n")\
             .append("\t")\
             .append("^\\s+$")
+        self.contains_article = './/article|.//*[@id="article"]|.//*[@itemprop="articleBody"]'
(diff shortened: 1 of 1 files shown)


newspaper/cleaners.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, config):
             .create("\n", "\n\n")\
             .append("\t")\
             .append("^\\s+$")
+        self.contains_article = './/article|.//*[@id="article"]|.//*[@itemprop="articleBody"]'
```
codelucas: nit: rename this to be `contains_article__re`? similar to the other variable names?

codelucas:
This seems like a great fix for little cost - good heuristic, thanks.

I wish we had a big testing suite of articles to validate successes / failures against, that way we can quantitatively determine the impact of these changes..

PR #425, opened 2017-08-30
Mark elements within a link as highlink_density.

megatron-me-uk:
If an element is wrapped in a link, it should be considered to have a high link density.


codelucas:
Thanks for putting in the effort on this @megatron-me-uk. Can you add a comment or even some unit tests showing why this change will improve text extraction accuracy?

codelucas:
I rethought this design and am not sure if it would help (unless we had specific cases where this code can improve text extraction). High-link density is a heuristic that reveals if a bunch of text is spammy and thus not part of the full-text. Not sure in which scenario the entire article would be wrapped up in the same <a>

Issue #379, opened 2017-06-10
TypeError: 'int' object does not support item assignment


fhamborg:
Running on Python 3.5 (Linux)

```
[...]
  File "/home/andreas/Envs/newsi/lib/python3.5/site-packages/newspaper/article.py", line 219, in parse
    meta_data = self.extractor.get_meta_data(self.clean_doc)
  File "/home/andreas/Envs/newsi/lib/python3.5/site-packages/newspaper/extractors.py", line 514, in get_meta_data
    ref[part] = value
TypeError: 'int' object does not support item assignment
```

codelucas:
Thanks for finding this!

codelucas:
Will investigate, there seems to be a few errors in the html parse/extract logic

codelucas:
The latest fix by @megatron-me-uk at https://github.com/codelucas/newspaper/pull/426 should fix your issue :) Please pull the latest master version of newspaper and reattempt this.

Otherwise, I plan on bumping the newspaper version and updating sometime this week.

PR #426, opened 2017-08-30
Treat metadata integers the same as strings

megatron-me-uk:
I have encountered an error 'in the wild' that was correctly handled by this change.

Should fix issue #379

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -504,7 +504,7 @@ class ContentExtractor(object):
             key_head = key.pop(0)
             ref = data[key_head]
 
-            if isinstance(ref, str):
+            if isinstance(ref, str) or isinstance(ref, int):
                 data[key_head] = {key_head: ref}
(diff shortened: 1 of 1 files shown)


codelucas:
Thanks for doing this @megatron-me-uk, will take a look and get back to the PR soon

codelucas:
Looks like a good fix, thanks so much for doing this 👍 💯 ! Will also close the related task as it does look like this solves the issue.

newspaper/extractors.py
```diff
@@ -504,7 +504,7 @@ def get_meta_data(self, doc):
             key_head = key.pop(0)
             ref = data[key_head]
 
-            if isinstance(ref, str):
+            if isinstance(ref, str) or isinstance(ref, int):
```
codelucas: I wonder if the `ref / ref = data[key_head]` object can be of any other type besides `str` or `int`? If a float appears the same bug would happen correct?

PR #427, opened 2017-08-31
extract chinese title problem

xiezhi546:
Currently, it can not extract chinese title in HTML. Because the 'get_title' method of extractors.py has a regular expression which not include chinese unicode, so need add it.  - from Gridsum


codelucas:
This is a great find @xiezhi546, also good solution to fix this 👍 ! I just double checked your commit to confirm it would work on a Chinese site. I am going to merge this into master now.

Expect your commit to make it into `pip` package manager in a few days.

PR #428, opened 2017-09-02
Add Ukrainian language support

older:
This adds support for Ukrainian language.
Ukrainian language uses Cyrillic and splitting/tokenizing using spaces works fine.


codelucas:
Thank you so much for doing this @older! Approving and merging now. 👍 

It will be a few days before this change makes it into newspaper's latest `pip` version :)

PR #430, opened 2017-09-04
Better get_meta_lang

Cabu:
Added one more case in the language detection.
To solve a problem with the following page: http://quebec.huffingtonpost.ca/2017/09/04/trump-evoque-la-vente-a-seoul-de-milliards-de-dollars-de-materiel-militaire_a_23196569/


codelucas:
Thanks for putting in the effort on this @Cabu, looking

PR #432, opened 2017-09-05
Fix attribute loss for div elements after running document cleaner

fengkaijia:
```div_to_para()``` uses ```doc.clear()``` to remove children nodes. However, while doing so it also removes attributes including id and class:

> Resets an element. This function removes all subelements, clears all attributes and sets the text and tail properties to None.

This PR backup the attributes and restore them after ```doc.clear()```.

### DIFF
--- a/newspaper/cleaners.py
+++ b/newspaper/cleaners.py
@@ -3,6 +3,7 @@
 Holds the code for cleaning out unwanted tags from the lxml
 dom xpath.
 """
+import copy
 from .utils import ReplaceSequence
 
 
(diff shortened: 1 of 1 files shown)


codelucas:
Good find! I am merging this into master 👍 💯 

Did you find that extraction quality improved a lot after fixing this?

fengkaijia:
@codelucas Rarely, I guess it mainly helps post-processing, i.e. reuses the ```self.doc``` to do other tasks, some sites loss a lot of id/class and become hard to do ```cssselect``` after cleaning.

codelucas:
@fengkaijia yep, in terms of correctness this is great, thanks a lot! And who knows, maybe in the future we will do more post-processing steps and this will help extraction quality

PR #438, opened 2017-09-13
Fix #118 Language-specific stopwords

### LINKED ISSUE #118: Set Language through Config() doesn't set language specific stopwords and Article's keyword problem.
Hi.

I've seen that there's a bug where when you are setting the config of the newspaper.build() it doesn't load the stopwords array for the desired language.
The nlp keywords function is predefined to load a stopwords file like this:

``` python
with open(settings.NLP_STOPWORDS_EN, 'r') as f:
    stopwords = set([w.strip() for w in f.readlines()])
```

And the function:

``` python
def keywords(text):
    """Get the top 10 keywords and their frequency scores ignores blacklisted
    words in stopwords, counts the number of occurrences of each word, and
    sorts them in reverse natural order (so descending) by number of
    occurrences.
    """
    NUM_KEYWORDS = 10
    text = split_words(text)
    # of words before removing blacklist words
    if text:
        num_words = len(text)
        text = [x for x in text if x not in stopwords]
        freq = {}
        for word in text:
            if word in freq:
                freq[word] += 1
            else:
                freq[word] = 1

        min_size = min(NUM_KEYWORDS, len(freq))
        keywords = sorted(freq.items(),
                          key=lambda x: (x[1], x[0]),
                          reverse=True)
        keywords = keywords[:min_size]
        keywords = dict((x, y) for x, y in keywords)

        for k in keywords:
            articleScore = keywords[k]*1.0 / max(num_words, 1)
            keywords[k] = articleScore * 1.5 + 1
        return dict(keywords)
    else:
        return dict()
```

So there's no difference in telling in the config class what language to use, because nlp() always calls the previous function which loads the English nlp (which it's in misc/ directory and it's different from the text/'s directory one) stopwords file.

Could you explain if this is done on purpose or there's really something wrong or to be implemented in language specific keyword making.

obulat:
Loads language-specific stop words before executing` Article.nlp()` call for creating more meaningful summaries or keyword lists in languages other than English.
Fixes #118 
Changed encoding of 'stopwords-ru.txt' so that the first line is read correctly.


codelucas:
Thanks for finding this - i'm looking now

obulat:
Tests fail because there are two different files for English stop words: 

```
STOPWORDS_DIR = os.path.join(PARENT_DIRECTORY, 'resources/text')

# NLP stopwords are != regular stopwords for now...
NLP_STOPWORDS_EN = os.path.join(
    PARENT_DIRECTORY, 'resources/misc/stopwords-nlp-en.txt')
```
'resources/misc/stopwords-nlp-en.txt' is used for tests now.

Also, the following function opens resource file without specifying encoding, which results in an error for languages other than English:
```
def mock_resource_with(filename, resource_type):
    """
    Mocks an HTTP request by pulling text from a pre-downloaded file
    """
    VALID_RESOURCES = ['html', 'txt']
    if resource_type not in VALID_RESOURCES:
        raise Exception('Mocked resource must be one of: %s' %
                        ', '.join(VALID_RESOURCES))
    subfolder = 'text' if resource_type == 'txt' else 'html'
    resource_path = os.path.join(TEST_DIR, "data/%s/%s.%s" %
                                 (subfolder, filename, resource_type))
    with open(resource_path, 'r') as f:
        return f.read()
```

codelucas:
This looks great @obulat, thanks for the fix! 👍 💯 Will merge and commit a new version to pip tonight! Looks like this was quite a big unfixed bug.

newspaper/nlp.py
```diff
@@ -9,17 +9,34 @@
...
+    """
+    global stopwords
+    
+    # stopwords for nlp in English are not the regular stopwords
+    # to pass the tests
+    # can be changed with the tests
+    if language == 'en':
+        stopwordsFile = settings.NLP_STOPWORDS_EN
+    else:
+        stopwordsFile = path.join(settings.STOPWORDS_DIR,\
```
codelucas: nit: `stopwordsFile` should be `stopwords_file` per python syntax, but letting this fly since your change is crucial.

newspaper/nlp.py
```diff
@@ -9,17 +9,34 @@
...
 from . import settings
 
-with open(settings.NLP_STOPWORDS_EN, 'r') as f:
-    stopwords = set([w.strip() for w in f.readlines()])
-
 ideal = 20.0
 
+stopwords = set()
 
+def load_stopwords(language):
```
codelucas: One interesting side-effect of you moving the stopwords loading code from the global scope into a function is that the stopwords loading is now going to be run dynamically when `Article.nlp()` is called .. instead of being run at import time.

I am not sure if this side-effect will consequences

newspaper/resources/text/stopwords-ru.txt
```diff
@@ -1,4 +1,4 @@
-﻿а
+а
```
codelucas: Thanks for this!

newspaper/nlp.py
codelucas: Actually, after merging your change I'm going to commit a change which moves all of the `load_stopwords` code into **global scope** and also deletes the `nlp.load_stopwords(self.config.get_language())` in article.py to minimize side effects.

codelucas:
💯 thanks for this again!

obulat:
Thank you for merging and kind comments! This is my first PR.

PR #441, opened 2017-09-15
Add test for prepare_url

fengkaijia:
This only test if URLs are combined correctly.


codelucas:
Fantastic work dude, thanks @fengkaijia 💯

PR #443, opened 2017-09-19
Allow to define extra headers on the configuration object

ametade:
If a dictionary with the headers is defined it will be used when doing requests.


codelucas:
Thanks for doing this, great work 💯 👍

ametade:
Thank you for accepting my PR!

PR #445, opened 2017-09-25
free optimizations

mercuree:
Less code, does the same, works much faster when intensively used with some big pages.

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -863,12 +863,7 @@ class ContentExtractor(object):
         return False
 
     def walk_siblings(self, node):
-        current_sibling = self.parser.previousSibling(node)
-        b = []
-        while current_sibling is not None:
-            b.append(current_sibling)
-            current_sibling = self.parser.previousSibling(current_sibling)
-        return b
+        return self.parser.previousSiblings(node)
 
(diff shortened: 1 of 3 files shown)


newspaper/extractors.py
```diff
@@ -863,12 +863,7 @@ def is_boostable(self, node):
         return False
 
     def walk_siblings(self, node):
-        current_sibling = self.parser.previousSibling(node)
-        b = []
-        while current_sibling is not None:
-            b.append(current_sibling)
-            current_sibling = self.parser.previousSibling(current_sibling)
-        return b
+        return self.parser.previousSiblings(node)
```
codelucas: Good optimization 1 liner :) 💯 👍

codelucas: I didn't even realize that this function was just re-using the previousSiblings code, wow maybe we can just replace the function directly in the future.

newspaper/parsers.py
```diff
@@ -211,28 +211,18 @@ def getText(cls, node):
 
     @classmethod
     def previousSiblings(cls, node):
```
codelucas: Confirmed the changes on this file are all proper via: http://lxml.de/api/lxml.etree._Element-class.html

PR #456, opened 2017-10-07
Convert list items correctly (fixes double text)

mercuree:
This fix is text-related only. 
Text within list items should appear entirely and without doubling.
I also changed test files (which were already buggy) for clarity.

I would suggest to prefix list items with [bullet sign](https://en.wikipedia.org/wiki/Bullet_(typography)) like `•` or `[*]`

### DIFF
--- a/newspaper/outputformatters.py
+++ b/newspaper/outputformatters.py
@@ -88,9 +88,7 @@ class OutputFormatter(object):
         for e in self.parser.getElementsByTag(self.top_node, tag='ul'):
             li_list = self.parser.getElementsByTag(e, tag='li')
             for li in li_list[:-1]:
-                li.text = self.parser.getText(li) + r'\n'
-                for c in self.parser.getChildren(li):
-                    self.parser.remove(c)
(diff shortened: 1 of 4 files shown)


newspaper/outputformatters.py
```diff
@@ -88,9 +88,7 @@ def add_newline_to_li(self):
         for e in self.parser.getElementsByTag(self.top_node, tag='ul'):
             li_list = self.parser.getElementsByTag(e, tag='li')
             for li in li_list[:-1]:
-                li.text = self.parser.getText(li) + r'\n'
-                for c in self.parser.getChildren(li):
```
codelucas: @mercuree Why did you also remove the `parser.getChildren` `remove` call?

mercuree: It will be converted to text anyway in a few calls:
https://github.com/mercuree/newspaper/blob/f73af8c7114f7623a1e20437c26adb647ce1ce26/newspaper/outputformatters.py#L59

The `remove` was called to avoid text-doubling, but it was not enough. 
We can just not touch node tree there, `convert_to_text` will do the job.

PR #466, opened 2017-10-30
change newspaper.build(url,limit=n) behavior

N2ITN:
The limit value set in `newspaper.build(url,limit=n)` will limit the number of articles downloaded by `source.Source.categories_to_articles()` and `source.Source.feeds_to_articles()`. it will stop when the number set is reached, as opposed to scraping for all articles and limiting the number of results returned.

This changes the runtime of the following from 9.6 seconds to 1.8 (>5 times faster).
```python
         src = newspaper.build(
                'http://cnn.com',
                fetch_images=False,
                request_timeout=2,
                limit=30,
                memoize_articles=False)
```
Note: My IDE did auto-formatting for style, so it looks like there are a lot more changes than I made. 

Sorry about that.

Thanks for your hard work, this library is awesome!


codelucas:
Thanks for doing this @N2ITN, will take a look. 👍 💯

natecorkish:
Any idea if this will be accepted?

PR #476, opened 2017-11-24
modify the last step of title-extracting

Ckins:
As stated in the original comments at line 343, we prefer the title_text_h1, when the filter_title is similar to the title_text_h1. However,  operator '==' in Python is too strict. I suggest using the string comparaing method in the library 'difflib' to judge wheter title_text h1 is simialr to the filter_title.


codelucas:
Cool proposal @Ckins! Will take a look, thanks so much for doing this 💯 👍

PR #480, opened 2017-12-01
Added reference corpus keyword functionality

IngoKl:
This pull request adds a new functionality to the NLP module that is widely used in corpus linguistics. The idea is to extracts keywords by statistically comparing a text (or corpus) to a reference corpus that is (potentially) representative of a language/genre. 

I implemented the functionality in a way that the user passes the reference corpus to the `nlp` method. I'm not perfectly clear whether this is the ideal way.

Thank you for considering!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,8 @@
 Newspaper3k: Article scraping & curation
 ========================================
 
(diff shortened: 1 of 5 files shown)


codelucas:
Interesting stuff @IngoKl! Thanks for doing this, will take a look when I have time 💯 👍

PR #488, opened 2017-12-13
Put error print inside a check for verbosity

SN9NV:
Checks if the verbosity flag is set before printing error.


codelucas:
This seems logical to do, not sure why it was not wrapped to begin with. Thoughts @yprez?

codelucas:
Will merge this soon. Thanks for adding this @SN9NV 👍

PR #491, opened 2017-12-22
Add a simple client

jaysonsantos:
Add as simple client in order to download articles and pipe to commands
like speedread `newspaper -u url | speedread` and read articles from the command line.


codelucas:
Interesting proposal, will look & think about merging this. Thanks for the good owrk @jaysonsantos!
👍 💯

codelucas:
Looks pretty good, doing some more testing on this before merging

PR #496, opened 2018-01-03
Image Fetch Rework for Performance

owaaa:
This is my work in an attempt to address #483.

The Image Fetch configuration now only controls whether images will be downloaded. If fetching images is false in the configuration, images will still be returned analyzed through the other methods (top node etc). It is possible for top_image to be `None` if it can not be determined w/o fetching.

Image fetch when true (left default) will still work as previous, except that image URLs may still be checked for validity based on the valid URL heuristics.

This also adds a new heuristic around valid image urls. It is written in a very similar way to regular link validity, just tuned initially for images.

This PR has been required for me to use newspaper on a large volume of URLs without certain urls causing blocks.

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -252,8 +252,7 @@ class Article(object):
             self.set_article_html(article_html)
             self.set_text(text)
 
-        if self.config.fetch_images:
-            self.fetch_images()
+        self.fetch_images()
 
         self.is_parsed = True
         self.release_resources()
@@ -272,9 +271,12 @@ class Article(object):
         if self.clean_top_node is not None and not self.has_top_image():
             first_img = self.extractor.get_first_img_url(
                 self.url, self.clean_top_node)
-            self.set_top_img(first_img)
+            if self.config.fetch_images:
+                self.set_top_img(first_img)
+            else:
+                self.set_top_img_no_check(first_img)
 
-        if not self.has_top_image():
+        if not self.has_top_image() and self.config.fetch_images:
             self.set_reddit_top_img()
 
     def has_top_image(self):
@@ -436,7 +438,8 @@ class Article(object):
 
     def set_meta_img(self, src_url):
         self.meta_img = src_url
-        self.set_top_img_no_check(src_url)
+        if images.valid_image_url(src_url):
+            self.set_top_img_no_check(src_url)
 
     def set_top_img(self, src_url):
         if src_url is not None:
--- a/newspaper/images.py
+++ b/newspaper/images.py
@@ -27,7 +27,7 @@ chunk_size = 1024
 thumbnail_size = 90, 90
 minimal_area = 5000
 
-BAD_URLS = ['logo', 'icon', 'banner', 'ribbon', 'asset', 'sprite', 'qrcode']
+BAD_URLS = ['logo', 'icon', 'banner', 'ribbon', 'sprite', 'qrcode']
 ALLOWED_TYPES = ['jpg', 'jpeg', 'png']
 
 def image_to_str(image):
@@ -189,7 +189,7 @@ def valid_image_url(url):
     # '/story/cnn/blahblah/index.html' --> ['story', 'cnn', 'blahblah', 'index.html']
     path_chunks = [x for x in path.split('/') if len(x) > 0]
 
-    # siphon out the file type. eg: .html, .htm, .md
+    # siphon out the file type. eg: jpeg, png
     if len(path_chunks) > 0:
         file_type = urls.url_to_filetype(url)
 
--- a/newspaper/images.py
+++ b/newspaper/images.py
@@ -13,6 +13,8 @@ import math
 import io
 import traceback
 import urllib.parse
+from urllib.parse import urlparse
+from tldextract import tldextract
 
 import requests
 from PIL import Image, ImageFile
@@ -25,6 +27,8 @@ chunk_size = 1024
 thumbnail_size = 90, 90
 minimal_area = 5000
 
+BAD_URLS = ['logo', 'icon', 'banner', 'ribbon', 'asset', 'sprite', 'qrcode']
+ALLOWED_TYPES = ['jpg', 'jpeg', 'png']
 
 def image_to_str(image):
     s = io.StringIO()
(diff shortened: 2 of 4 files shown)


newspaper/images.py
```diff
@@ -25,6 +27,8 @@
 thumbnail_size = 90, 90
 minimal_area = 5000
 
+BAD_URLS = ['logo', 'icon', 'banner', 'ribbon', 'sprite', 'qrcode']
```
owaaa: This is based on my own research, I tried to be conservative to avoid stripping out interesting images. I could see making this configurable but it is not currently on the other valid_url checks

newspaper/images.py
```diff
@@ -25,6 +27,8 @@
...
+ALLOWED_TYPES = ['jpg', 'jpeg', 'png']
```
owaaa: should .gif be allowed? normally its associated with logos or regular page assets

newspaper/article.py
```diff
@@ -272,9 +271,12 @@ def fetch_images(self):
...
             first_img = self.extractor.get_first_img_url(
                 self.url, self.clean_top_node)
-            self.set_top_img(first_img)
+            if self.config.fetch_images:
+                self.set_top_img(first_img)
+            else:
+                self.set_top_img_no_check(first_img)
 
-        if not self.has_top_image():
+        if not self.has_top_image() and self.config.fetch_images:
```
owaaa: the meat of the fix, where most time is spent on some sites.

codelucas:
Thanks for taking this on @owaaa, will take a look at the PR 💯 👍

newspaper/images.py
vitaly-zdanevich: webp?

newspaper/article.py
codelucas: 👍

newspaper/images.py
```diff
@@ -167,6 +171,52 @@ def fetch_image_dimension(url, useragent, referer=None, retries=1):
...
+
+    # '/story/cnn/blahblah/index.html' --> ['story', 'cnn', 'blahblah', 'index.html']
+    path_chunks = [x for x in path.split('/') if len(x) > 0]
+
+    # siphon out the file type. eg: jpeg, png
+    if len(path_chunks) > 0:
+        file_type = urls.url_to_filetype(url)
+
+        # if the file type is a media type, reject instantly
+        if file_type and file_type not in ALLOWED_TYPES:
```
codelucas: This is good because if the file_type is "" -> if the URL resource does not specify the file suffix we also don't aggressively reject the file.

newspaper/images.py
codelucas: Let's leave gifs out for now - good call. Can we be careful about what is in ALLOWED_TYPES, maybe also add TIFF, EPS, and RAW? Also maybe webp like vitaly suggested above, I'm not sure what image files are popular nowadays.
http://guides.lib.umich.edu/c.php?g=282942&p=1885348

newspaper/images.py
```diff
@@ -167,6 +171,52 @@ def fetch_image_dimension(url, useragent, referer=None, retries=1):
     return fetch_url(url, useragent, referer, retries, dimension=True)
 
 
+def valid_image_url(url):
```
codelucas: This is great work! But it's also used as a core function, maybe we can add a few simple unit tests for this function on say 10 good image URLs and 10 bad image urls just so we have sanity knowing it works?

newspaper/images.py
```diff
@@ -167,6 +171,52 @@ def fetch_image_dimension(url, useragent, referer=None, retries=1):
...
+
+    url = clean_url(url)
+    if not url.startswith(('http://', 'https://')):
```
codelucas: But what if the input URL can be a relative URL?

newspaper/images.py
```diff
@@ -167,6 +171,52 @@ def fetch_image_dimension(url, useragent, referer=None, retries=1):
...
+        return False
+    path = urlparse(url).path
+
+    # input url is not in valid form (scheme, netloc, tld)
```
codelucas: Same as above, we don't know the expectation of this function, the input url may be relative, e.g. `/images/content/red_car.jpg`

newspaper/images.py
```diff
@@ -167,6 +171,52 @@ def fetch_image_dimension(url, useragent, referer=None, retries=1):
...
+        # the file type is not of use to use anymore, remove from url
+        if len(last_chunk) > 1:
+            path_chunks[-1] = last_chunk[-2]
+
+
+    # extract the tld (top level domain)
+    tld_dat = tldextract.extract(url)
+    tld = tld_dat.domain.lower()
+
+    if tld in urls.BAD_DOMAINS:
```
codelucas: Nice job re-using the BAD_DOMAINS list 👍

owaaa:
@codelucas I split out #511 as requested. Thanks for the other great feedback, I'll come back to that when I get some more cycles

codelucas:
#511 is merged, 👌

newspaper/images.py
owaaa: from my reading I don't think RAW or EPS are supported in browsers. There are website that purport to allow viewing of those. 
https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support
WebP seems good. TIFF seems rarely supported but we could add it

newspaper/images.py
owaaa: will do

newspaper/images.py
owaaa: good call

newspaper/images.py
owaaa: added, see below

newspaper/article.py
```diff
@@ -438,7 +438,8 @@ def set_article_html(self, article_html):
 
     def set_meta_img(self, src_url):
         self.meta_img = src_url
-        self.set_top_img_no_check(src_url)
+        if images.valid_image_url(src_url):
```
owaaa: here is a question, check this or always trust meta?

owaaa:
@codelucas think i'm done for now with updates unless you guys have additional feedback. thanks

owaaa:
following up. Anything else needed? I've been running with a fork tied to this since Jan and seems ok

PR #498, opened 2018-01-04
Fix wrong format of debug message in generate_articles method

korycins:


codelucas:
LGTM! 👍 💯 thanks for doing this @korycins

PR #504, opened 2018-01-12
Add encoding inference.

zhaodaolimeng:
Requests library used by Newspaper strictly follows http RFC, which prevent Newspaper from parsing many 
Chinese website without encoding in response head.

These code fix this issue using standard method provided by Requests library.

### DIFF
--- a/newspaper/network.py
+++ b/newspaper/network.py
@@ -80,8 +80,13 @@ def _get_html_from_response(response):
         # return response as a unicode string
         html = response.text
     else:
-        # don't attempt decode, return response in bytes
         html = response.content
+        if 'charset' not in response.headers.get('content-type'):
+            encodings = requests.utils.get_encodings_from_content(response.text)
+            if len(encodings) > 0:
+                response.encoding = encodings[0]
+                html = response.text
+
     return html or ''
 
 


codelucas:
Thanks so much for finding this @zhaodaolimeng. 💯 👍  let me quickly search and make sure this is the recommended way of handling responses without given encodings

codelucas:
Hmm, in requests official documentation here: http://docs.python-requests.org/en/master/user/advanced/#encodings

It states that if charset is not present in content-type it defaults to the ISO-8859-1 charset. Why is this behavior worse than using `requests.utils.get_encodings_from_content`, do most non-english websites use a different encoding?

codelucas:
Actually after reading the `get_encodings_from_content`, I think your PR has the best solution! 👍  My only comment is that the `if 'charset' not in response.headers.get('content-type'):` guard is not needed?

newspaper/network.py
```diff
@@ -80,8 +80,13 @@ def _get_html_from_response(response):
         # return response as a unicode string
         html = response.text
     else:
-        # don't attempt decode, return response in bytes
         html = response.content
+        if 'charset' not in response.headers.get('content-type'):
```
codelucas: Is this `if` statement needed? Since at this stage `response.encoding` is already empty or failed, we want to run `requests.utils.get_encodings_from_content` no matter what.

I think we should remove the if check and just keep
```
html = response.content
encodings = requests.utils.get_encodings_from_content(response.text)
if len(encodings) > 0:
  response.encoding = encodings[0]
  html = response.text
```

zhaodaolimeng: Actually, I'm not sure, but suppose not.  This `if` is added because when the encoding is ISO-8895-1(which is the major case), the whole branch will be run, and I think the `get_encodings_from_content` maybe less efficient(a regular expression match?).

codelucas: I see, thanks

PR #506, opened 2018-01-17
Update ReadMe

nnamdiib:
The example used in the README file, http://cnn.com does not actually work. 
CNN now keeps their news articles at http://edition.cnn.com .
If you type cnn.com into a browser, you get redirected to edition.cnn.com

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,7 +78,7 @@ A Glance:
 
     >>> import newspaper
 
-    >>> cnn_paper = newspaper.build('http://cnn.com')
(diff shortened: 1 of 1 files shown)


codelucas:
@nnamdiib thanks for finding this! 👍  however I am having trouble replicating, I am NOT redirected to edition.cnn.com from both typing in cnn.com into the browser and also clicking on any headline..

Which browser are you using and which country are you testing this from? Maybe that's related?

PR #511, opened 2018-01-24
Don't fetch images but still return images if configured to not fetch…

owaaa:
… images

Split PR as requested in PR feedback for #496


codelucas:
LGTM 💯

PR #523, opened 2018-02-26
added estonian stopwords

Florian-Roessler:
added Estonian stop words based on the following resources:

https://github.com/voku/stop-words/blob/master/src/voku/helper/stopwords/et.php
https://github.com/6/stopwords-json/blob/master/dist/et.json
https://github.com/stopwords-iso/stopwords-et/blob/master/stopwords-et.txt

Would appreciate if a someone more familiar with the language could double check these.


codelucas:
Thanks for adding this @Florian-Roessler! It is especially helpful since you cited where you got the stopwords. Merging 👍 💯

PR #524, opened 2018-02-28
Fixed problem with empty meta description

baktakt:
I have found examples where sites use the syntax <meta property="description" ... 

This makes the get_meta_description return "". Added check to see if the doc uses the property syntax instead. See for example: https://www.sydsvenskan.se/2018-02-28/200-malmopoliser-har-fatt-kroppskameror-pa-forsok

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -475,7 +475,10 @@ class ContentExtractor(object):
     def get_meta_description(self, doc):
         """If the article has meta description set in the source, use that
         """
-        return self.get_meta_content(doc, "meta[name=description]")
+        if len(self.get_meta_content(doc, "meta[name=description]")) > len(self.get_meta_content(doc, "meta[property=description]")):
+          return self.get_meta_content(doc, "meta[name=description]")
+      
+        return self.get_meta_content(doc, "meta[property=description]")
 
     def get_meta_keywords(self, doc):
         """If the article has meta keywords set in the source, use that


codelucas:
Thanks so much for creating this, looking soon! 👍

newspaper/extractors.py
```diff
@@ -475,7 +475,10 @@ def get_meta_type(self, doc):
     def get_meta_description(self, doc):
         """If the article has meta description set in the source, use that
         """
-        return self.get_meta_content(doc, "meta[name=description]")
+        if len(self.get_meta_content(doc, "meta[name=description]")) > len(self.get_meta_content(doc, "meta[property=description]")):
```
codelucas: This adds unnecessary cpu cycles to parsing, since we call `self.get_meta_content(doc, "meta[name=description]")` multiple times when we only need to call it once.

How about this:
```
meta_name_text = self.get_meta_content(doc, "meta[name=description]")
meta_property_text = self.get_meta_content(doc, "meta[property=description]")
if len(meta_name_text) > len(meta_property_text):
  return meta_name_text
return meta_property_text
```

Or even shorter:
```
return self.get_meta_content(doc, "meta[name=description]") or self.get_meta_content(doc, "meta[property=description]")
```

codelucas (changes requested):
This is a great fix but please see inline comments for efficiency increases before approval/merging.

Thanks! 💯

newspaper/extractors.py
codelucas: Also, as a design question, can it ever be the case that the *shorter* meta description is the correct description?

PR #532, opened 2018-03-15
Allow configuration of base directory

prevostc:
I was able to make newspaper work on aws lambda with this PR update and the following code

```
# only /tmp is writable on lambda env
os.environ['NEWSPAPER_BASE_DIRECTORY'] = '/tmp'
os.environ['NLTK_DATA'] = '/tmp'

# https://stackoverflow.com/a/44532317/2523414
# no sqlite in lambda, needed by a small subset of nltk
import imp
import sys
sys.modules["sqlite"] = imp.new_module("sqlite")
sys.modules["sqlite3.dbapi2"] = imp.new_module("sqlite.dbapi2")

from newspaper import Article
import nltk
nltk.download('punkt')
```

This PR is also handy for AWS Beanstalk as the `/home/wsgi` folder do not exists and you also need to configure the directory

### DIFF
--- a/newspaper/settings.py
+++ b/newspaper/settings.py
@@ -34,7 +34,9 @@ DATA_DIRECTORY = '.newspaper_scraper'
 
 
 if 'LAMBDA_TASK_ROOT' in os.environ:
-    BASE_DIRECTORY = os.environ['LAMBDA_TASK_ROOT']
+    # https://docs.aws.amazon.com/lambda/latest/dg/limits.html
+    # only /tmp is writeable in lambda
+    BASE_DIRECTORY = "/tmp"
 else:
     BASE_DIRECTORY = tempfile.gettempdir()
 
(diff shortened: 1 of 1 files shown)


codelucas:
Great idea, thanks for implementing this @prevostc. Will take a look / test and likely merge in a few days! 👍 💯

chingan-tsc:
@prevostc Thanks for the much-needed PR for Newspaper to work AWS Lambda, I am trying the same thing as you did and its working quite well. However I noticed its rather slow even with 512MB of RAM, I was wondering do you have any optimization ideas on this?

torbenbrodt:
+1

prevostc:
This PR is now 1.5 year old, which is a billion serverless years. I'm closing it :)

Nop0x:
Why did this not get merged? This is really needed in my opinion.

vchalana1:
Hello, Is this change in the 0.2.8 release? I did a pip install of the latest release as of March 2, 2020, and I don't see any NEWSPAPER_BASE_DIRECTORY.

PR #544, opened 2018-04-03
Nepali language support

ashokpant:
Added Nepali stopwords.


newspaper/configuration.py
```diff
@@ -5,6 +5,7 @@
 holds them. For example, pass in a config object to an Article
 object, Source object, or even network methods, and it just works.
 """
+
```
codelucas: Please remote added emptyline

newspaper/text.py
```diff
@@ -182,3 +182,29 @@ def get_stopword_count(self, content):
         ws.set_stopword_count(len(overlapping_stopwords))
         ws.set_stop_words(overlapping_stopwords)
         return ws
+
+
+class StopWordsNepali(StopWords):
```
codelucas: Notice that `StopWordsNepali` has identical code with the inherited `Stopwords`. Please correct me if wrong but we don't need the new `StopWordsNepali` class, we can just re-use `Stopwords` for the `np` language

newspaper/configuration.py
```diff
@@ -108,6 +109,8 @@ def get_stopwords_class(language):
             return StopWordsKorean
         elif language == 'hi':
             return StopWordsHindi
+        elif language == 'np':
```
codelucas: Remove these two lines, we can just use `Stopwords` generically for `np`

codelucas (changes requested):
Thanks so much for adding this change @ashokpant 👍 💯 Please see my inline comments and review and fix before we can merge :)

ashokpant:
Hi @codelucas, @yprez 
Could you guys please review it again? Let me know if I have to correct anything for merging the PR.

PR #548, opened 2018-04-09
Enable async downloading of individually specified articles

lebedov:
The docstring of `newspaper.mthreading.NewsPool.__init__` appears to indicate that one can pass it a list of source or article objects. The class implementation, however, cannot handle article objects; I modified `newspaper.mthreading.NewsPool.set` to remedy this.

### DIFF
--- a/newspaper/mthreading.py
+++ b/newspaper/mthreading.py
@@ -15,7 +15,6 @@ from threading import Thread
 
 from .configuration import Configuration
 
-
 class Worker(Thread):
     """
     Thread executing tasks from a given tasks queue.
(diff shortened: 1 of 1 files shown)


codelucas:
Hey @lebedov, this is an awesome fix! Will merge now 👍 💯

newspaper/mthreading.py
```diff
@@ -105,4 +104,7 @@ def set(self, paper_list, threads_per_source=1):
         self.pool = ThreadPool(num_threads, timeout)
 
         for paper in self.papers:
-            self.pool.add_task(paper.download_articles)
+            try:
+                self.pool.add_task(paper.download_articles)
+            except AttributeError:
```
codelucas: I guess long-term the clean solution would be to see `isinstance(object, Article) or isinstance(object, Source)` and call the appropriate method after. I'll add a TODO

PR #550, opened 2018-04-12
Fixed raise statement

ttong1013:
Quick fix to the raise statement in mthreading.py

### DIFF
--- a/newspaper/configuration.py
+++ b/newspaper/configuration.py
@@ -39,7 +39,10 @@ class Configuration(object):
         self.MAX_FILE_MEMO = 20000
 
         # Cache and save articles run after run
-        self.memoize_articles = True
+        self.memoize_articles = False
+
+        # Limit of number of articles
+        self.limit = 10000
 
         # Set this to false if you don't care about getting images
         self.fetch_images = True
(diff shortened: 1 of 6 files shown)


newspaper/configuration.py
```diff
@@ -39,7 +39,10 @@ def __init__(self):
         self.MAX_FILE_MEMO = 20000
 
         # Cache and save articles run after run
-        self.memoize_articles = True
+        self.memoize_articles = False
```
codelucas: Can you revert these changes?

newspaper/configuration.py
```diff
@@ -39,7 +39,10 @@ def __init__(self):
...
+
+        # Limit of number of articles
+        self.limit = 10000
```
codelucas: ditto

newspaper/version.py
```diff
@@ -7,5 +7,5 @@
 __license__ = 'MIT'
 __copyright__ = 'Copyright 2014, Lucas Ou-Yang'
 
-version_info = (0, 2, 6)
+version_info = (0, 2, 7)
```
codelucas: Why bump the version?

codelucas (changes requested):
Thanks for this workl @ttong1013! But I think you left some unrelated changes in this commit/PR

PR #553, opened 2018-04-17
Convert mthreading to use concurrent.futures

c0d3d:
to provide support for futures and async processing

This came about when I wanted to be able to process jobs as they were completed, but the only supported operations for the thread pool was to submit new jobs, and to wait for all the jobs to complete. This left the main thread blocked unnecessarily.

### DIFF
--- a/newspaper/mthreading.py
+++ b/newspaper/mthreading.py
@@ -9,50 +9,58 @@ __author__ = 'Lucas Ou-Yang'
 __license__ = 'MIT'
 __copyright__ = 'Copyright 2014, Lucas Ou-Yang'
 
-import queue
-import traceback
-from threading import Thread
+from concurrent.futures import ThreadPoolExecutor, wait
+from threading import Lock
 
 from .configuration import Configuration
 
-
-class Worker(Thread):
-    """
-    Thread executing tasks from a given tasks queue.
-    """
-    def __init__(self, tasks, timeout_seconds):
-        Thread.__init__(self)
-        self.tasks = tasks
-        self.timeout = timeout_seconds
-        self.daemon = True
-        self.start()
-
-    def run(self):
-        while True:
-            try:
-                func, args, kargs = self.tasks.get(timeout=self.timeout)
-            except queue.Empty:
-                # Extra thread allocated, no job, exit gracefully
-                break
-            try:
-                func(*args, **kargs)
-            except Exception:
-                traceback.print_exc()
-
-            self.tasks.task_done()
-
+def perform_download_articles(paper):
+    return paper.download_articles
 
 class ThreadPool:
-    def __init__(self, num_threads, timeout_seconds):
-        self.tasks = queue.Queue(num_threads)
-        for _ in range(num_threads):
-            Worker(self.tasks, timeout_seconds)
+    def __init__(self, num_threads, timeout_seconds=0):
+        self.mutex = Lock()
+        self.executor = ThreadPoolExecutor(max_workers=num_threads)
+        self.timeout_seconds = timeout_seconds
+        self.futures = {}
 
     def add_task(self, func, *args, **kargs):
-        self.tasks.put((func, args, kargs))
+        """
+        Adds the given task, and returns a future for the completion.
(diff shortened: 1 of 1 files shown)


newspaper/mthreading.py
```diff
@@ -9,50 +9,58 @@
...
     def add_task(self, func, *args, **kargs):
-        self.tasks.put((func, args, kargs))
+        """
+        Adds the given task, and returns a future for the completion.
+        """
+        f = self.executor.submit(func, *args, **kargs)
+
+        f.add_done_callback(self._remove_future)
+
+        self.mutex.acquire()
```
maciej-gol:
```python
with self.mutex:
    self.futures[f] = 1
```
?

c0d3d: Sorry, didn't know you could do that. Much better ...

codelucas:
Thanks so much for this change @c0d3d, will take some time to review 👍 💯

newspaper/mthreading.py
```diff
@@ -81,28 +80,29 @@ def __init__(self, config=None):
...
-        self.papers = paper_list
-        num_threads = threads_per_source * len(self.papers)
-        timeout = self.config.thread_timeout_seconds
-        self.pool = ThreadPool(num_threads, timeout)
+        # Wait for completion of all futures.
+        self.pool.wait_completion()
 
-        for paper in self.papers:
-            self.pool.add_task(paper.download_articles)
+    def set(self, paper_list, num_threads=8, fun=perform_download_articles):
```
codelucas: This changes control behavior, it used to be 1 thread per source but with this change it will be 8 threads no matter what. I'm scared existing callers using this API will break after this change:

Can we revert this bit to go back to `threads_per_source`, later I can add another API to allow num_threads to be configured because we are moving to a world where `set` can be called on Source objects and also Article objects

newspaper/mthreading.py
```diff
@@ -9,50 +9,49 @@
...
-                break
-            try:
-                func(*args, **kargs)
-            except Exception:
-                traceback.print_exc()
-
-            self.tasks.task_done()
-
+def perform_download_articles(paper):
+    return paper.download_articles
```
codelucas: This behavior has since been changed since your PR, can you replace this line with:

```
from newspaper.source import Source
...
if isinstance(paper, Source):
  return paper.download_articles
else:
  return paper.download
```

newspaper/mthreading.py
```diff
@@ -9,50 +9,49 @@
...
 
 class ThreadPool:
-    def __init__(self, num_threads, timeout_seconds):
-        self.tasks = queue.Queue(num_threads)
-        for _ in range(num_threads):
-            Worker(self.tasks, timeout_seconds)
+    def __init__(self, num_threads, timeout_seconds=0):
```
codelucas: This changes control behavior drastically, before `timeout_seconds` was given from our Config object at `self.config.thread_timeout_seconds`, now it is a keyword arg defaulting to 0. Can we remove the default from this param and properly set timeout to `self.config.thread_timeout_seconds` like before?

codelucas (changes requested):
Looks good! But I found a few instances where control behavior was changed, can you look at inline comments and fix.

Also, before we merge this change, do you have performance metrics (like for example how fast the lib runs on the old api vs. this new concurrent one you've added on 1000 article downloads? we want to make sure there is no regression)

Thanks again for this great change

newspaper/mthreading.py
c0d3d: I understand the concert, but I think that the number of threads shouldn't effect callers that used to use the old version. The behavior of the old pool only let the caller wait until all the sources were done, so how they get to that point shouldn't matter. I'm still going to make the change to make the API the same as it used to be though.

codelucas: 👌

hugocool:
Is there any update on this merge request? I would love to use this functionality!

c0d3d:
@hugocool I never got around to making the requested changes or benchmarks, so now progress.

PR #561, opened 2018-05-07
Made the amount of keywords adjustable, defaulting to 10

fabiant7t:
The maximum number of keywords was hard coded to be 10. To be able to use the keywords method directly and make it return more than 10 keywords, I added the `num_keywords` argument that defaults to 10.


codelucas:
@fabiant7t The two changes that you've made in this are great, but can u address this first:

You have fixed the German stopwords and also cleaned up the license, great! But can you split that logic into a separate pull request and name it accordingly (this one is named for the keyword change) The stopword and keyword changes are very different so it's important for them to be in separate PRs.

newspaper/nlp.py
```diff
@@ -116,13 +116,12 @@ def split_words(text):
         return None
 
 
-def keywords(text):
+def keywords(text, num_keywords=10):
```
codelucas: This is an improvement! but can you go even further and make this configurable via the config file?
https://github.com/codelucas/newspaper/blob/624a6ea77d55575f77bd1f6697034dc575b05c7d/newspaper/configuration.py#L23

PR #564, opened 2018-05-09
Added support for multi-threaded Article downloading

Briscoooe:
In the `mthreading.py` class, the NewsPool object description is as follows:

```
Abstraction of a threadpool. A newspool can accept any number of
source OR article objects together in a list. It allocates one
thread to every source and then joins.
```

However, I noticed it could not support the the inclusion of `Article` objects because the method passed to `pool.add_task` is `Source.download_articles()`. When passing a list of `Article`s, this this line throws an exception.

What I've done is added some minor, non breaking changes to support multithreaded `Article` downloading, identical to how it can be done with `Source`s. Below is a summary of changes:
- [Added a `set_articles` method and `articles` property to the `NewsPool` class](https://github.com/codelucas/newspaper/compare/master...Briscoooe:master#diff-125b00b6b6bf2170179699d51bfbfd8c)
- [Added a `with_nlp` argument to the `Article.build()` method with a default value of `True`, so as not to break any existing code](https://github.com/codelucas/newspaper/compare/master...Briscoooe:master#diff-58c6e0d178f563a62a6450c669930681)
- [Added a max number of threads to the default config](https://github.com/codelucas/newspaper/compare/master...Briscoooe:master#diff-8c4b8ecb8a8e68e766479760763fa8d0)

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -143,15 +143,16 @@ class Article(object):
         # A property dict for users to store custom data.
         self.additional_data = {}
 
-    def build(self):
+    def build(self, with_nlp=True):
         """Build a lone article from a URL independent of the source (newspaper).
         Don't normally call this method b/c it's good to multithread articles
         on a source (newspaper) level.
         """
         self.download()
         self.parse()
-        self.nlp()
-
+        if with_nlp:
+            self.nlp()
+    
     def download(self, input_html=None, title=None, recursion_counter=0):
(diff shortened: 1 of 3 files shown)


c0d3d:
See: #553 Would be cool to integrate these two :P

I guess we need some maintainer response first ...

Briscoooe:
Any updates on this?

PR #570, opened 2018-05-21
Allow .shtml pages

### LINKED ISSUE #569: Allow `.shtml` pages
`.shtml` is currently not an allowed type for [URLs](https://github.com/codelucas/newspaper/blob/c521057b20bb3d4cd27d8b0ee6efd64d1d3a488f/newspaper/urls.py#L27), while some newspaper (e.g.: http://www.ilsole24ore.com/) still use it, therefore the `newspaper.build()` fail to parse some articles

rpadovani:
Fix #569


codelucas:
This is a great catch, thank you @rpadovani! 💯 👍

PR #573, opened 2018-05-31
Fix get_html_2XX_only exceptions masking

pyontko-nazar:
get_html_2XX_only in network.py masks all exceptions, so it was no possibilities to get the real error message in "outside code".

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -520,17 +520,14 @@ class Article(object):
         -> maybe throw ArticleException
         """
         if self.download_state == ArticleDownloadState.NOT_STARTED:
-            print('You must `download()` an article first!')
-            raise ArticleException()
+            raise ArticleException('You must `download()` an article first!')
         elif self.download_state == ArticleDownloadState.FAILED_RESPONSE:
-            print('Article `download()` failed with %s on URL %s' %
+            raise ArticleException('Article `download()` failed with %s on URL %s' %
                   (self.download_exception_msg, self.url))
-            raise ArticleException()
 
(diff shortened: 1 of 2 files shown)


newspaper/network.py
```diff
@@ -59,12 +59,8 @@ def get_html_2XX_only(url, config=None, response=None):
     if response is not None:
         return _get_html_from_response(response)
 
-    try:
-        response = requests.get(
-            url=url, **get_request_kwargs(timeout, useragent, proxies, headers))
-    except requests.exceptions.RequestException as e:
```
codelucas: Good catch, this try / catch handler is not needed since the call sites seem to all have the same handler anyways..

newspaper/article.py
```diff
@@ -520,17 +520,14 @@ def throw_if_not_downloaded_verbose(self):
         -> maybe throw ArticleException
         """
         if self.download_state == ArticleDownloadState.NOT_STARTED:
-            print('You must `download()` an article first!')
```
codelucas: The print statements instead of text embedded into a custom thrown exception was bad design from my part at the start. Thanks for this.

We should get in a habit of migrating all these print statements into custom exceptions or `LOG` messages at least.

newspaper/network.py
codelucas: However FYI that the design in the html fetch functions of catch exceptions is intentional because otherwise when we are downloading 1000+ articles from CNN for example, if 100 articles gave a timeout error it would make more sense to catch those errors and print the messages instead of throw an individual uncaught exception for the first failed article and kill the whole flow

PR #575, opened 2018-06-08
Scalable code for more websites

Ask149:
I have been using this library for a project in hand. And I found that your library could not get authors and publishing date for some of the website's articles. I would strongly suggest you accept the pull request to scale it to get the author names from more websites. I will soon try to come up with a solution for the publishing date.


newspaper/extractors.py
```diff
@@ -148,7 +148,7 @@ def parse_byline(search_str):
 
         for match in matches:
             content = ''
-            if match.tag == 'meta':
+            if match.tag :
```
codelucas: Removing the meta filter will reduce the performance of all `parse_byline` calls because instead of just the few `<meta>` headers we permit, we'd be running xpath searches on all tags with `ATTRS` and `VALS` combinations. If you have in mind a list of candidates (in addition to just the `meta` tag) we can consider, let's add that as an `if match.tag in set([x, y, z]):` check instead?

What tags outside of meta did you notice were being left out?

Ask149: Sure I'll get you with a list of those tags and make a pull request soon

PR #584, opened 2018-06-26
Added Japanese language support.

sfi-dannybrady:
I've added Japanese language support. Japanese needs a new tokenizer, so [tinysegmenter](https://pypi.org/project/tinysegmenter/) was used. The language code for Japanese is ja.


requirements.txt
```diff
@@ -10,3 +10,4 @@ tldextract>=2.0.1
 feedfinder2>=0.0.4
 jieba3k>=0.35.1
 python-dateutil>=2.5.3
+tinysegmenter==0.3
```
codelucas: Do you mean `>=0.3` ? Let's merge anyways but add a todo to change this if needed

codelucas:
Will deploy this in pypi after some more testing on my end

PR #586, opened 2018-06-27
Added auto language detection when language isn't in meta data

sfi-dannybrady:
If the meta data language can't be found, and the language wasn't provided by the user, then try to detect the language of the article from the title text. If the language detected is in the list of available languages, use it, if it's not in the list don't use it.

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -10,6 +10,7 @@ import os
 import glob
 
 import requests
+from langdetect import detect
 
 from . import images
 from . import network
@@ -210,6 +211,14 @@ class Article(object):
         if self.config.use_meta_language:
             self.extractor.update_language(self.meta_lang)
             output_formatter.update_language(self.meta_lang)
+
+            # If there is no meta language provide, but detect_lang_from_title is true, and there is title text, try to detect language
+            if not self.meta_lang and self.config.detect_lang_from_title and self.title:
+                detected_lang = detect(self.title)
+                # Make sure the detected language is an available language, or don't use it
+                if detected_lang in get_available_languages():
+                    self.extractor.update_language(detected_lang)
+                    output_formatter.update_language(detected_lang)
 
         meta_favicon = self.extractor.get_favicon(self.clean_doc)
         self.set_meta_favicon(meta_favicon)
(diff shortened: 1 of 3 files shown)


newspaper/article.py
```diff
@@ -10,6 +10,7 @@
 import glob
 
 import requests
+from langdetect import detect
```
codelucas: `detect` is easy to overload, can we `import langdetect` and use `langdetect.detect(..)`?

newspaper/article.py
```diff
@@ -211,6 +212,14 @@ def parse(self):
             self.extractor.update_language(self.meta_lang)
             output_formatter.update_language(self.meta_lang)
 
+            # If there is no meta language provide, but detect_lang_from_title is true, and there is title text, try to detect language
+            if not self.meta_lang and self.config.detect_lang_from_title and self.title:
+                detected_lang = detect(self.title)
```
codelucas: I would rather use the more low-level API of getting probabilities instead of just picking the highest probability language.

For example:
what if the title was incredibly ambiguous and the "top lang detection" was only 10% sure? versus a long and obvious title that gets a 99% detection of English?

langdetect has a `get_probabilities(..)` API that allows fetching the direct probabilities. Let us grab the highest probability in 0th index and only use the detected language if it passes our threshold instead?
```
probs = langdetect.get_probabilities(self.title)
if len(probs) > 0 and probs[0].prob > SOME_THRESHOLD_PROBABLY_90_PERCENT:
  if probs[0].lang in get_available_languages():
      ... process
```
https://github.com/Mimino666/langdetect/blob/master/langdetect/detector.py#L141-L144

Does that work?

codelucas (commented):
This is great idea and implementation @sfi-dannybrady, thanks 👍 💯 ! Before merging let me first verify a few things with the dependency you've added https://github.com/Mimino666/langdetect.

Want to make sure this dep is 1) small and straightforward to install, since newspaper already has a lot of heavy deps we don't want to add more unless it unlocks crucial features and 2) does langdetect have a reasonable accuracy

Also added some inline comments. Definitely interested in adding this feature since lang detection is crucial :) maybe we can even expand detection to the "top paragraph" of the article

PR #597, opened 2018-07-19
Update stopwords-pt.txt

eduardogoncalves:
Update list with pt stopwords from: https://github.com/stopwords-iso/stopwords-pt;
Sort words alphabetically.


codelucas:
Merging anyways as stopwords-iso seems to be a trustworthy source of stopwords, but ideally in the future we should have a system to give us metrics of stopwords performance so when we make changes we know it is actually improving extraction ability

thanks again!

PR #598, opened 2018-07-20
Handle file scheme in Article.download

dlrobertson:
- Update the Article download function to handle the file scheme.
 - Add test cases for using Article.download with a file url

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -8,6 +8,7 @@ import logging
 import copy
 import os
 import glob
+from urllib.parse import urlparse
 
 import requests
 
@@ -152,6 +153,23 @@ class Article(object):
         self.parse()
         self.nlp()
 
+    def _parse_scheme_file(self, path):
+        try:
+            with open(path, "r") as fin:
+                return fin.read()
+        except OSError as e:
+            self.download_state = ArticleDownloadState.FAILED_RESPONSE
+            self.download_exception_msg = e.strerror
+            return None
+
+    def _parse_scheme_http(self):
+        try:
+            return network.get_html_2XX_only(self.url, self.config)
+        except requests.exceptions.RequestException as e:
+            self.download_state = ArticleDownloadState.FAILED_RESPONSE
+            self.download_exception_msg = str(e)
+            return None
+
     def download(self, input_html=None, title=None, recursion_counter=0):
         """Downloads the link's HTML content, don't use if you are batch async
         downloading articles
@@ -160,11 +178,12 @@ class Article(object):
         infinite
         """
         if input_html is None:
-            try:
-                html = network.get_html_2XX_only(self.url, self.config)
-            except requests.exceptions.RequestException as e:
-                self.download_state = ArticleDownloadState.FAILED_RESPONSE
-                self.download_exception_msg = str(e)
+            parsed_url = urlparse(self.url)
+            if parsed_url.scheme == "file":
+                html = self._parse_scheme_file(parsed_url.path)
+            else:
+                html = self._parse_scheme_http()
+            if html is None:
                 log.debug('Download failed on URL %s because of %s' %
                           (self.url, self.download_exception_msg))
                 return
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -25,6 +25,7 @@ URLS_FILE = os.path.join(TEST_DIR, 'data', 'fulltext_url_list.txt')
 
 import newspaper
 from newspaper import Article, fulltext, Source, ArticleException, news_pool
+from newspaper.article import ArticleDownloadState
 from newspaper.configuration import Configuration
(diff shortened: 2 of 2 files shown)


tests/unit_tests.py
```diff
@@ -322,6 +325,26 @@ def test_nlp_body(self):
...
+    def test_download_file_success(self):
+        url = "file://" + os.path.join(HTML_FN, "cnn_article.html")
+        article = Article(url=url)
+        article.download()
+        self.assertEqual(article.download_state, ArticleDownloadState.SUCCESS)
+        self.assertEqual(article.download_exception_msg, None)
+        self.assertEqual(75406, len(article.html))
+
+    @print_test
+    def test_download_file_failure(self):
```
codelucas: Thanks for the tests

newspaper/article.py
```diff
@@ -160,14 +161,26 @@ def download(self, input_html=None, title=None, recursion_counter=0):
...
-            try:
-                html = network.get_html_2XX_only(self.url, self.config)
-            except requests.exceptions.RequestException as e:
-                self.download_state = ArticleDownloadState.FAILED_RESPONSE
-                self.download_exception_msg = str(e)
-                log.debug('Download failed on URL %s because of %s' %
-                          (self.url, self.download_exception_msg))
-                return
+            parsed_url = urlparse(self.url)
+            if parsed_url.scheme == "file":
```
codelucas: Now that the `def download(..)` fn is quite large let's split up the logic of handling `if parsed_url.scheme == "file":` versus scheme http in different helper functions?

Maybe something two new functions in the `Article`  class such as
```
def _parse_scheme_file(parsed_url):
  try:
    with open(parsed_url.path, "r") as fin:
      return fin.read()
  except OSError as e:
    self.download_state = ArticleDownloadState.FAILED_RESPONSE
    self.download_exception_msg = e.strerror
    log.debug('Download failed on URL %s because of %s' %
                      (self.url, self.download_exception_msg))
    return

def _parse_scheme_http():
  try:
    html = network.get_html_2XX_only(self.url, self.config)
  except requests.exceptions.RequestException as e:
    self.download_state = ArticleDownloadState.FAILED_RESPONSE
    ... <and so on>

def download(...):
  ... <bunch of logic>

  if parsed_url.scheme == "file":
    _parse_scheme_file(parsed_url)
  else:
    _parse_scheme_http()

```

newspaper/article.py
```diff
@@ -160,14 +161,26 @@ def download(self, input_html=None, title=None, recursion_counter=0):
...
+                    self.download_state = ArticleDownloadState.FAILED_RESPONSE
+                    self.download_exception_msg = e.strerror
+                    log.debug('Download failed on URL %s because of %s' %
+                              (self.url, self.download_exception_msg))
+                    return
+            else:
+                try:
+                    html = network.get_html_2XX_only(self.url, self.config)
+                except requests.exceptions.RequestException as e:
+                    self.download_state = ArticleDownloadState.FAILED_RESPONSE
```
codelucas: alternatively maybe we can have a way to share logic for the error handing, since the error handing of both file and http failing are quite similar

dlrobertson:
Updated. Sorry for the delay in updating the PR.

newspaper/article.py
```diff
@@ -152,6 +153,23 @@ def build(self):
         self.parse()
         self.nlp()
 
+    def _parse_scheme_file(self, path):
```
codelucas: @dlrobertson what are the use-cases you saw where you'd need to parse `file://` resources instead of the typical `https://`, `http://`, and so on? Did you want to use this lib to crawl offline file-systems of html pages?

I am wondering if this use-case is common

newspaper/article.py
```diff
@@ -160,14 +178,14 @@ def download(self, input_html=None, title=None, recursion_counter=0):
...
-                self.download_state = ArticleDownloadState.FAILED_RESPONSE
-                self.download_exception_msg = str(e)
+            parsed_url = urlparse(self.url)
+            if parsed_url.scheme == "file":
+                html = self._parse_scheme_file(parsed_url.path)
+            else:
+                html = self._parse_scheme_http()
+            if html is None:
                 log.debug('Download failed on URL %s because of %s' %
                           (self.url, self.download_exception_msg))
```
codelucas: I will approve after you add back this `return` statement, as if html is none we can skip logic the further post-processing, the return gives a bit of clarity to control flow
Nice job!!

newspaper/article.py
dlrobertson: > Did you want to use this lib to crawl offline file-systems of html pages? 

Yes

> I am wondering if this use-case is common.

Probably not too common.

newspaper/article.py
dlrobertson: Updated

PR #599, opened 2018-07-23
Add publish_time_telegram to publish_date tags

dlrobertson:
Add the `publish_time_telegram` tag to the list of possible tags containing the published date.

```python
from newspaper import Article

article = Article(url="https://www.rt.com/uk/434030-sharks-british-waters-devon/")
article.download()
article.parse()
print(article.publish_date)
```

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -216,6 +216,8 @@ class ContentExtractor(object):
              'content': 'content'},
             {'attribute': 'pubdate', 'value': 'pubdate',
              'content': 'datetime'},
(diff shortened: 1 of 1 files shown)


codelucas:
Nice find, thank you @dlrobertson. Do you know how often `publish_time_telegram` is used in news sites (outside of the particular case of rt.com?) I am close to merging / accepting this change but since it does slightly increase extraction time I want to confirm `publish_time_telegram` is a standard that some news sites follow

dlrobertson:
It doesn't seem to be used too often. Probably a better solution would be to add something to the `Configuration` class.

PR #627, opened 2018-09-17
Added div_to_para transformation for section tags

NunoPinheiro:
Section tags are pretty interchangeable with div tags, so I believe giving them the same processing that is given to div tags makes sense.


codelucas:
@NunoPinheiro very cool find 💯 , [this stackoverflow link](https://stackoverflow.com/questions/6939864/what-is-the-difference-between-section-and-div/6941170#6941170) makes me inclined to accept your PR immediately since it seems to be important to capture `<section>`s. Nice work, this is like one of those 1-liners that has big impact

codelucas:
Merging now, @NunoPinheiro can you also post here some empirical proof where you found that cleaning sections makes for better extractions?

Thanks

PR #630, opened 2018-10-10
Add extraction of meta_site_name

pgrzesik:
Hello,
we found out that very often it's useful to extract not only source_url, but also site_name (from Open Graph) protocol. While this is available from meta_data, we thought that it might be convinent to have it as a property on the article itself and following PR adds that functionality.

Regards,
Piotr


newspaper/outputformatters.py
```diff
@@ -141,9 +141,12 @@ def remove_empty_tags(self):
     def remove_trailing_media_div(self):
         """Punish the *last top level* node in the top_node if it's
         DOM depth is too deep. Many media non-content links are
-        eliminated: "related", "loading gallery", etc
+        eliminated: "related", "loading gallery", etc. It skips removal if
+        last top level node's class is one of NON_MEDIA_CLASSES.
         """
 
+        NON_MEDIA_CLASSES = ('zn-body__read-all', )
```
codelucas: I merged your other commit, can you rebase so these changes don't show?

pgrzesik: sure thing - changed!

Issue #632, opened 2018-10-11
CI setup not running `fulltext` tests


pgrzesik:
Hello,

while working on one of the PRs, I noticed that current setup of CI is not running `fulltext` tests, which might result in them failing on master branch (current case, described in: https://github.com/codelucas/newspaper/issues/631). I will gladly help with adjusting configuration of Travis so it will also run fulltext tests.

Issue #631, opened 2018-10-11
Fulltext tests failing on current master


pgrzesik:
Hello,

while working on one of the PRs, I noticed that current master branch has failing `fulltext` tests. Let me know if I can help in resolving that issue.

Issue #560, opened 2018-05-06
CNN articles are not correctly parsed


jensdebruijn:
Parsing CNN links yields Read More in the article text.

```
url = "http://www.cnn.com/2017/01/30/politics/trump-immigration-ban-refugees-trnd/index.html"
article = Article(url)
article.download()
article.parse()
print(article.text)
```

yields

`Being a refugee isn't easy. Neither is getting into the United States.                                                                                                                                                                                                                                                                                                                                                                                                                                      Just gaining refugee status is hard; most of the millions who seek it don't get it. Of those who do, only a fraction end up resettled in the US.                                                                                                                                                                                                                                                                                                                                                            The process takes an average of 18 to 24 months, and most of it's conducted outside the US.                                                                                                                                                                                                                                                                                                                                                                                                                 Here's how it works:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        It's a long and winding road                                                                                                                                                                                                                                                                                                                                                                                                                                                                                Read More`

pgrzesik:
This PR is solving this issue: https://github.com/codelucas/newspaper/pull/633

PR #633, opened 2018-10-11
Skip removing last diff it it's known that it has non-media class

pgrzesik:
Hello,

while trying to handle articles from CNN, we stumbled across issue with skipping part of the article, after investigation it turned out that it was because of too agressive removal of last div, which is adjusted in the following PR. Keep in mind that this is only a proposition of implementation and I would love to discuss other possible approaches. I can confirm that this solves following issue: https://github.com/codelucas/newspaper/issues/560

Example articles that had this issue:
https://edition.cnn.com/2017/01/30/politics/trump-immigration-ban-refugees-trnd/index.html
https://edition.cnn.com/2018/08/21/us/colorado-chris-watts-court/index.html

Moreover, during testing I noticed that current master has failing `fulltext` tests which are also not run on Travis, which has been filled in form of following issues:
https://github.com/codelucas/newspaper/issues/632
https://github.com/codelucas/newspaper/issues/631

Let me know if I can help with fixing them/improving CI setup, I would love to contribute here.

Regards,
Piotr

### DIFF
--- a/newspaper/outputformatters.py
+++ b/newspaper/outputformatters.py
@@ -141,8 +141,11 @@ class OutputFormatter(object):
     def remove_trailing_media_div(self):
         """Punish the *last top level* node in the top_node if it's
         DOM depth is too deep. Many media non-content links are
-        eliminated: "related", "loading gallery", etc
+        eliminated: "related", "loading gallery", etc. It skips removal if
+        last top level node's class is one of NON_MEDIA_CLASSES.
         """
+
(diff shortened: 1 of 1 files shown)


codelucas:
Thanks so much for the effort in finding / addressing this issue @pgrzesik 👍 💯 Especially since news traffic's distribution is head heavy, fixing issues with CNN gives very large impact

codelucas:
Approved and merging

PR #635, opened 2018-10-12
Included 'span' and 'a'

agnelvishal:
Sometimes author names are included inside span and a. So modified the code. 
Can also include div if necessary

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -148,7 +148,7 @@ class ContentExtractor(object):
 
         for match in matches:
             content = ''
-            if match.tag == 'meta':
+            if match.tag == 'meta' or match.tag == 'span' or match.tag == 'a':
                 mm = match.xpath('@content')
                 if len(mm) > 0:
                     content = mm[0]


codelucas:
Great find @agnelvishal, looking!! 💯👍

newspaper/extractors.py
```diff
@@ -148,8 +148,10 @@ def parse_byline(search_str):
 
         for match in matches:
             content = ''
-            if match.tag == 'meta':
+            if match.tag == 'meta' or match.tag == 'span' or match.tag == 'a':
                 mm = match.xpath('@content')
+                if not mm:
+                    mm=str(match.text_content()).split()
```
codelucas: Confident that this will increase the recall but will precision take a hit?

I am going to approve this but can you also add some urls where you found that these changes were helpful

codelucas: nit: python style, please add spaces between the ` = ` on line 154

agnelvishal:
Could modify https://github.com/codelucas/newspaper/blob/b75b9c75195cc45e198075f578ed9ce7d10c5b6a/newspaper/extractors.py#L159 as  ` if len(content) > 0 and len(content) < 30:` to get better precision.

PR #636, opened 2018-10-12
Added "Tamil" Language Stop words

agnelvishal:


codelucas:
Thanks for doing this @agnelvishal! This travis CI build is failing, I suspect it has to do with Tamil language code being 3 characters instead of the expected 2.

codelucas:
I restarted your build, let's see what happens

codelucas:
The problem is that your stopwords code is 3 character ISO-639-2 instead of the 2 character variant.
See Tamil 2 char variants: https://en.wikipedia.org/wiki/Tamil_language

All our current stopwords are in 2 char form and we have unit tests that error out if they are in 3 char form: https://github.com/codelucas/newspaper/tree/master/newspaper/resources/text

PR #650, opened 2018-11-09
Updated date tags

agnelvishal:
Websites like https://www.thehindu.com/business/Industry/netflix-unveils-asia-expansion-plans-with-17-new-originals/article25451041.ece have these tags

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -216,6 +216,12 @@ class ContentExtractor(object):
              'content': 'content'},
             {'attribute': 'pubdate', 'value': 'pubdate',
              'content': 'datetime'},
+            {'attribute': 'name', 'value': 'publish-date',
+             'content': 'content'},
+            {'attribute': 'name', 'value': 'created-date',
+             'content': 'content'},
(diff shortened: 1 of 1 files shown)


codelucas:
thanks for adding this @agnelvishal this looks great. 👍 💯  I want to first do some research to make sure the 3 tags you added are also used for a wide variety of other sites as well before merging. otherwise additional checks will slow down the codepath for all news sites

newspaper/extractors.py
```diff
@@ -216,6 +216,12 @@ def parse_date_str(date_str):
              'content': 'content'},
             {'attribute': 'pubdate', 'value': 'pubdate',
              'content': 'datetime'},
+            {'attribute': 'name', 'value': 'publish-date',
+             'content': 'content'},
+            {'attribute': 'name', 'value': 'created-date',
+             'content': 'content'},
+            {'attribute': 'name', 'value': 'modified-date',
```
Kerl1310: I'd be careful with modified-date - I've recently been stung where the true publication date was early 2000s whereas the modified date was a recent date due to updating copyright or some other unrelated content

PR #651, opened 2018-11-12
Fix extracting proper author information with nested tags

torbenbrodt:
tested with https://ekstrabladet.dk/plus/sportogspil/stor-oversigt-over-superliga-traenernes-loen-staale-spraenger-banken/7392243?icname=plus-front&icpos=63_1&iccreative=7392243

where author is in nested structure
```
<span itemprop="author" itemscope itemtype="http://schema.org/Person">
                        <span itemprop="name">
                                Klaus Egelund</span>
                    </span>
```

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -153,7 +153,7 @@ class ContentExtractor(object):
                 if len(mm) > 0:
                     content = mm[0]
             else:
-                content = match.text or ''
+                content = match.text_content() or ''
             if len(content) > 0:
                 authors.extend(parse_byline(content))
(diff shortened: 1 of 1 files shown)


newspaper/extractors.py
```diff
@@ -153,7 +153,7 @@ def parse_byline(search_str):
                 if len(mm) > 0:
                     content = mm[0]
             else:
-                content = match.text or ''
+                content = match.text_content() or ''
```
codelucas: This is a great idea, simple fix that may have high impact.

PR #655, opened 2018-11-21
Patch 1

agnelvishal:
The pull request code provides author name and date with high accuracy for good number of sites. It uses json-ld. More about json-ld at https://developers.google.com/search/docs/guides/intro-structured-data and https://stackoverflow.com/questions/38670851/whats-a-script-type-application-ldjsonjsonobj-script-in-a-head-sec

Also article date can be extracted from json-ld. As requested by @codelucas in #385, extruct need not be loaded

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -229,7 +229,7 @@ class Article(object):
         title = self.extractor.get_title(self.clean_doc)
         self.set_title(title)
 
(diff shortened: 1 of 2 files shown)


codelucas:
Very cool work @agnelvishal, thanks for doing this 👍 💯 I will look when free

PR #658, opened 2018-11-27
Ignoring http pages depending on their content-type

ekaterinasmarp:
When I run the download of long pdf article it often get failed by timeout on my application side. 

Change: 
- PDFs are ignored by default
- It's possible to add other content types for the ignoring list
- It's possible to switch off this ignoring by passing configuration ignored_content_types_defaults value

### DIFF
--- a/newspaper/configuration.py
+++ b/newspaper/configuration.py
@@ -34,6 +34,7 @@ class Configuration(object):
         self.MAX_AUTHORS = 10  # num strings in list
         self.MAX_SUMMARY = 5000  # num of chars
         self.MAX_SUMMARY_SENT = 5  # num of sentences
+        self.EMPTY_PDF = "%PDF-"  # empty PDF constant
 
         # max number of urls we cache for each news source
         self.MAX_FILE_MEMO = 20000
@@ -73,6 +74,10 @@ class Configuration(object):
 
         self.thread_timeout_seconds = 1
 
+        self.ignored_content_types_defaults = {"application/pdf": self.EMPTY_PDF,
+                                               "application/x-pdf": self.EMPTY_PDF,
+                                               "application/x-bzpdf": self.EMPTY_PDF,
+                                               "application/x-gzpdf": self.EMPTY_PDF}
         # Set this to False if you want to recompute the categories
         # *every* time you build a `Source` object
         # TODO: Actually make this work
--- a/newspaper/configuration.py
+++ b/newspaper/configuration.py
@@ -34,7 +34,6 @@ class Configuration(object):
         self.MAX_AUTHORS = 10  # num strings in list
         self.MAX_SUMMARY = 5000  # num of chars
         self.MAX_SUMMARY_SENT = 5  # num of sentences
-        self.EMPTY_PDF = "%PDF-"  # empty PDF constant
 
         # max number of urls we cache for each news source
         self.MAX_FILE_MEMO = 20000
@@ -73,11 +72,7 @@ class Configuration(object):
         self.verbose = False  # for debugging
 
         self.thread_timeout_seconds = 1
-
-        self.ignored_content_types_defaults = {"application/pdf": self.EMPTY_PDF,
-                                               "application/x-pdf": self.EMPTY_PDF,
-                                               "application/x-bzpdf": self.EMPTY_PDF,
-                                               "application/x-gzpdf": self.EMPTY_PDF}
+        self.ignored_content_types_defaults = {}
         # Set this to False if you want to recompute the categories
         # *every* time you build a `Source` object
         # TODO: Actually make this work
--- a/newspaper/network.py
+++ b/newspaper/network.py
@@ -57,12 +57,12 @@ def get_html_2XX_only(url, config=None, response=None):
     headers = config.headers
 
     if response is not None:
-        return _get_html_from_response(response)
+        return _get_html_from_response(response, config)
 
     response = requests.get(
         url=url, **get_request_kwargs(timeout, useragent, proxies, headers))
 
-    html = _get_html_from_response(response)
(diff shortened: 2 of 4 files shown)


newspaper/network.py
```diff
@@ -71,7 +71,10 @@ def get_html_2XX_only(url, config=None, response=None):
     return html
 
 
-def _get_html_from_response(response):
+def _get_html_from_response(response, config=None):
+    for disabled_type in config.ignored_content_types_defaults:
+        if disabled_type in response.headers.get('content-type'):
+            return config.ignored_content_types_defaults[disabled_type]
```
codelucas: Why return the sentinel empty pdf constant `%PDF-` instead of just an empty string? I feel returning special values like this will confuse the API, especially for a low-level function such as `_get_html_from_response`

newspaper/configuration.py
```diff
@@ -34,6 +34,7 @@ def __init__(self):
         self.MAX_AUTHORS = 10  # num strings in list
         self.MAX_SUMMARY = 5000  # num of chars
         self.MAX_SUMMARY_SENT = 5  # num of sentences
+        self.EMPTY_PDF = "%PDF-"  # empty PDF constant
```
codelucas: ditto: see comment below, I don't think having a constant sentinel for empty pdfs is a good idea

newspaper/configuration.py
```diff
@@ -73,6 +74,10 @@ def __init__(self):
 
         self.thread_timeout_seconds = 1
 
+        self.ignored_content_types_defaults = {"application/pdf": self.EMPTY_PDF,
```
codelucas: Perhaps make `self.ignored_content_types_defaults` a set of `set('application/pdf', application/x-pdf', ...)` since we don't need EMPTY_PDF anymore

newspaper/network.py
```diff
@@ -71,7 +71,10 @@ def get_html_2XX_only(url, config=None, response=None):
     return html
 
 
-def _get_html_from_response(response):
+def _get_html_from_response(response, config=None):
+    for disabled_type in config.ignored_content_types_defaults:
+        if disabled_type in response.headers.get('content-type'):
```
codelucas: As per the requests library API, `response.headers.get('content-type')` returns a string that represents the content-type. 

```
>>> resp = requests.get('https://www.adobe.com/pdf/pdfs/ISO32000-1PublicPatentLicense.pdf')
>>> resp.headers.get('content-type')
'application/pdf'
```

Why perform a substring match with `in` instead of just string comparison via `==`? Also this is not ideal for runtime complexity since you are checking every ignored content type in a loop, instead you can perform this instant lookup:

```
if response.headers.get('content-type') in config.ignored_content_types_defaults:
  return ''
```

newspaper/network.py
```diff
@@ -57,12 +57,12 @@ def get_html_2XX_only(url, config=None, response=None):
     headers = config.headers
 
     if response is not None:
-        return _get_html_from_response(response)
+        return _get_html_from_response(response, config)
 
     response = requests.get(
```
codelucas: Upon second thought, this is the line of code you should be worried about for the PDF timeouts, not the `_get_html_from_response()` function because at that point in code the long timeout for the PDF resource has already happened.

We should move the content type check logic before this line

tests/unit_tests.py
```diff
@@ -737,6 +737,23 @@ def test_languages_api_call(self):
         newspaper.languages()
 
 
+class TestDownloadPdf(unittest.TestCase):
```
codelucas: Thank you for these unit tests

codelucas:
Final request: can you also add instructions on how to add content-type filters to our library's users in the README or/and in the docs?

Here is an example of adding changes to our docs: https://github.com/codelucas/newspaper/pull/302/files
Here is a link to our docs: https://newspaper.readthedocs.io/en/latest/

newspaper/configuration.py
codelucas: Also, upon second thought I feel `self. ignored_content_types_defaults` should default to `{}`. The reason is it's not conventional / reasonable to change an API's default behaviors after many people have already downloaded and have expectations from it.

Like if we had someone using this library extensively and they expect PDFs to be returned but suddenly after the update their PDFs are all gone. It can cause a lot of headache.

newspaper/network.py
ekaterinasmarp: Thanks for your feedback, I agree this piece of code looks quite unexpected and ugli. Unfortunately I didn't found any standard way to understand what was the content type of the article from the client of the library. So what we are doing as a client of the library now is just checking the first 5 letters to be equal to this string. Is there any better way of getting the type of the article?

newspaper/network.py
ekaterinasmarp: Fixed

newspaper/network.py
```diff
@@ -71,7 +71,9 @@ def get_html_2XX_only(url, config=None, response=None):
     return html
 
 
-def _get_html_from_response(response):
+def _get_html_from_response(response, config=None):
```
codelucas: final nit: can you change `config=None` to just config, making this parameter optional is dangerous because we don't check if `config` is None, for example in line 75, we directly access `config.ignored_content_types_defaults` which will be a null pointer if someone accidentally calls `get_html_from_response` without the config param

Since this function is only used twice in the codebase, in both instances you have added the config param in the callsite, I think it's fine to force config to be a required param

codelucas:
Also, as a follow up TODO after this is merged it would be amazing if we added a guide (and suggestion for people on how to ignore PDFs or other big files) in the readme [tutorial on doing this](https://github.com/codelucas/newspaper/pull/302/files)

codelucas:
Excellent work @ekaterinasmarp! merging 💯

PR #661, opened 2018-12-29
Add more tags in extractor to search for authors

Ask149:
Hello @codelucas, I have made the changes you asked in my previous pull request ( 575 )

I am using the newspaper3k library for my project and I have noticed that it does not extract the list of authors from some websites. I have discovered a few tags which can contain the name of authors of articles. It would be really great if you could add those tags for searching authors.


Ask149:
Could we add this?

PR #693, opened 2019-03-31
Adding MIME type to Article

### LINKED ISSUE #576: Add mime_type to the Article object
For some scenarios, I have a need to actually see what is the MIME type of the resource `Newspaper` has downloaded. Since it is using `requests` to do the HTTP communication, this should not be hard to add. I can do a PR if maintainer (@codelucas) is OK with having this capability in the library.

thesmithee:
This PR adds MIME type to the Article class from the server. For provided HTML content it defaults to `text/html`. Similar for the `file:///` loaded articles. The latter is not strictly a correct behavior, rather we should try to deduce based on some set of well-known file extensions or similar mechanism. However, decided to put up the PR for discussion. 😃 

@codelucas can you PTAL and comment? 

Fixes #576

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -193,8 +193,12 @@ class Article(object):
             parsed_url = urlparse(self.url)
             if parsed_url.scheme == "file":
                 html = self._parse_scheme_file(parsed_url.path)
+                # Hard code to text/html for now
+                mime_type = "text/html"
             else:
-                html = self._parse_scheme_http()
+                result = self._parse_scheme_http()
+                html = result.html
+                mime_type = result.mime_type
             if html is None:
                 log.debug('Download failed on URL %s because of %s' %
                           (self.url, self.download_exception_msg))
@@ -204,12 +208,12 @@ class Article(object):
             # For now, just add a literal MIME type if html is set
             mime_type = "text/html"
 
-
         if self.config.follow_meta_refresh:
             meta_refresh_url = extract_meta_refresh(html)
             if meta_refresh_url and recursion_counter < 1:
                 return self.download(
                     input_html=network.get_html(meta_refresh_url).html,
+                    mime_type=mime_type,
                     recursion_counter=recursion_counter + 1)
 
         self.set_html(html)
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -153,6 +153,9 @@ class Article(object):
         # A property dict for users to store custom data.
         self.additional_data = {}
 
+        # A property to store MIME type that will be retrieved from network
+        self.mime_type = ""
+
     def build(self):
         """Build a lone article from a URL independent of the source (newspaper).
         Don't normally call this method b/c it's good to multithread articles
@@ -179,7 +182,7 @@ class Article(object):
             self.download_exception_msg = str(e)
             return None
 
-    def download(self, input_html=None, title=None, recursion_counter=0):
+    def download(self, input_html=None, title=None, recursion_counter=0, mime_type=None):
         """Downloads the link's HTML content, don't use if you are batch async
         downloading articles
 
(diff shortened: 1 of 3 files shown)


newspaper/network.py
```diff
@@ -68,7 +68,13 @@ def get_html_2XX_only(url, config=None, response=None):
         # fail if HTTP sends a non 2XX response
         response.raise_for_status()
 
-    return html
+    # Check to see if the Content-Type header exists
+    mime_type = None
+    if 'content-type' in response.headers:
+        mime_type = response.headers['content-type']
```
Kerl1310: This isn't necessarily correct. Content-Type is often made up of a MIME type combined with a charset like the one below:
`text/html; charset=utf-8`

newspaper/article.py
```diff
@@ -190,24 +193,32 @@ def download(self, input_html=None, title=None, recursion_counter=0):
             parsed_url = urlparse(self.url)
             if parsed_url.scheme == "file":
                 html = self._parse_scheme_file(parsed_url.path)
+                # Hard code to text/html for now
+                mime_type = "text/html"
```
Kerl1310: Not sure I agree with hard coding this to "text/html"... Personally I would probably rather lean towards None or an empty string

Kerl1310 (changes requested):

newspaper/article.py
blackdwarf: Yes, as mentioned in the PR, I don't think think this is correct either. If we don't want to be "smart" here, e.g. detect it somehow, then I guess `None` would be a good choice. Thoughts?

newspaper/network.py
blackdwarf: Based on [`Content-Type` docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) this seems like the right solution. Suggestions for a better header to take?

blackdwarf: One option would be to parse out stuff to `;` and take only the first part.

Kerl1310: I think the header is  the correct one to take, but if we're setting mime_type and not content_type then I'd do something like `mime_type = response.headers.get('content-type').split(';')[0]`

newspaper/article.py
Kerl1310: I'd lean towards `None`, simply because I've seen too many web pages that are PDFs but do not have the ".pdf" extension. This is just one example off the top of my head but I'm sure there are others

newspaper/network.py
blackdwarf: OK, will do, that makes sense. :)

newspaper/article.py
blackdwarf: `None` it is. :)

Kerl1310 (approved):

blackdwarf:
Hm, Travis is failing on Python 3.4, but it is weird, since I didn't touch any dependencies. Is this known issue?

Kerl1310:
Looks like someone has broken the build, not sure what the reason is. Maybe @louyangsc can advise?

iwpnd:
```
[...]
Collecting lxml>=3.6.0 (from -r requirements.txt (line 6))

  Downloading https://files.pythonhosted.org/packages/e1/f5/5eb3b491958dcfdcfa5daae3c655ab59276bc216ca015e44743c9c220e9e/lxml-4.4.0.tar.gz (4.5MB)

Complete output from command python setup.py egg_info:

    This lxml version requires Python 2.7, 3.5 or later.
```

`lxml` in version `4.4.0` dropped python 3.4 support, see [here](https://github.com/lxml/lxml/blob/lxml-4.4/setup.py). So @codelucas / you would have to fix an lxml version or drop python 3.4 support.

Kerl1310:
@iwpnd Think I've fixed the build issues in #768 if you could review?

PR #734, opened 2019-09-09
add new stopwords fa

mahekan:


mahekan:
add new stop word fa

Kerl1310:
@owaaa Could you review as you added the previous commit?

mahekan:
> @owaaa Could you review as you added the previous commit?

yes

PR #738, opened 2019-09-22
OutputFormatter - don't remove `img` tags

iAnanich:
I was working on a small project it matters to keep the original structure of an article.
And as I experienced so far, keeping `img` html tags in that way doesn't affect parsing efficiency.


iAnanich:
checks failed due to lxml compatibility issues with Python 3.4
(which can't be caused by my tiny change)

PR #768, opened 2020-01-14
Dropping python 3.4 support

### LINKED ISSUE #733: Python3.4 deprecated?
From https://travis-ci.org/codelucas/newspaper/builds/575679321 that was associated with a PR I made, the Python3.4 build failed, due to the reason:
```
This lxml version requires Python 2.7, 3.5 or later.
```

Kerl1310:
Fixes #733

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,4 +12,3 @@ PyYAML>=3.11
 requests>=2.10.0
 tinysegmenter==0.3  # TODO(codelucas): Investigate making this >=0.3
 tldextract>=2.0.1
-pandas>=0.25.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ PyYAML>=3.11
 requests>=2.10.0
 tinysegmenter==0.3  # TODO(codelucas): Investigate making this >=0.3
 tldextract>=2.0.1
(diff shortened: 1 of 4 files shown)


Kerl1310:
Fixes build issues

Kerl1310:
@codelucas Could you take a look at this PR? Fixes the issues with the build which would allow other changes to be made

requirements.txt
```diff
@@ -12,3 +12,4 @@ PyYAML>=3.11
 requests>=2.10.0
 tinysegmenter==0.3  # TODO(codelucas): Investigate making this >=0.3
 tldextract>=2.0.1
+pandas>=0.25.3
```
codelucas, line 15: why do this?

Kerl1310: Mistakenly added, shall remove momentarily

requirements.txt
```diff
@@ -11,4 +11,4 @@ python-dateutil>=2.5.3
 PyYAML>=3.11
 requests>=2.10.0
 tinysegmenter==0.3  # TODO(codelucas): Investigate making this >=0.3
-tldextract>=2.0.1
```
codelucas, line 14: i wonder why this was marked as a diff? since there is no diff

PR #781, opened 2020-02-15
Integration of YAKE!

JMendes1995:
According the email with Ricardo Campos. I submit the pull request with the integration of Key word extractor YAKE! with running examples in readme.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: python
 python:
- - "3.4"
  - "3.5"
  - "3.6"
+ - "3.7"
 install:
  - pip install -r requirements.txt coverage coveralls
  - python download_corpora.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+ - "3.4"
  - "3.5"
  - "3.6"
  - "3.7"
--- a/README.rst
+++ b/README.rst
@@ -77,7 +77,7 @@ A Glance:
 
 Keyword extraction with YAKE! (Yet Another keyword extractor)
 To learn more about YAKE! please check the official repository (https://github.com/LIAAD/yake).
-
+note: The lower the score, the more relevant the keyword is.
 .. code-block:: pycon
 
     >>> url = 'http://www.cnn.com/2013/11/27/justice/tucson-arizona-captive-girls/'
@@ -95,7 +95,7 @@ To learn more about YAKE! please check the official repository (https://github.c
     [('police', 0.02171325705124914), ('sophia richter', 0.02178687039012946),
     ('police chief roberto', 0.022941575782713476),...]
     
-    note: The lower the score, the more relevant the keyword is.
+    
     
 .. code-block:: pycon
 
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -12,3 +12,4 @@ PyYAML>=3.11
 requests>=2.10.0
 tinysegmenter==0.3  # TODO(codelucas): Investigate making this >=0.3
 tldextract>=2.0.1
+git+https://github.com/LIAAD/yake.git
```
codelucas, line 15: I normally would prefer you make this versioned like the above requirements. but this is OK since it seems the PR owners are all the creators of YAKE.

Is there a reason why this isn't a versioned package from PYPI?

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: python
 python:
- - "3.4"
```
codelucas, line 3: why remove 3.4? can u add it back

.travis.yml
```diff
@@ -1,8 +1,8 @@
...
  - "3.5"
  - "3.6"
+ - "3.7"
```
codelucas, line 5: 👌

README.rst
```diff
@@ -74,6 +74,29 @@ A Glance:
     >>> article.summary
     'The study shows that 93% of people ...'
 
+
+Keyword extraction with YAKE! (Yet Another keyword extractor)
+To learn more about YAKE! please check the official repository (https://github.com/LIAAD/yake).
+note: The lower the score, the more relevant the keyword is.
```
codelucas, line 80: curious, what does this score refer to?

newspaper/article.py
```diff
@@ -565,3 +565,9 @@ def throw_if_not_parsed_verbose(self):
         """
         if not self.is_parsed:
             raise ArticleException('You must `parse()` an article first!')
+
+    def yake(self, lan="en", n=3, dedupLim=0.9, dedupFunc='seqm', windowsSize=1, top=20, features=None):
+        import yake
```
codelucas, line 570: I think in-function imports are normally not preferred in python but in this case it's probably better incase yake installation fails we won't break the whole library, just this function call

README.rst
codelucas: part of me feels we should score=(1-score) so bigger is better. otherwise users of this API might get confused and try to sort(..) incorrectly.

you can also rename the api call `yake` to `yake_keywords_ranked(..)` so that it is clearer.

codelucas (changes requested):
this is great, apologies for the late review.

can u look over my comments/lints I plan to approve your PR shortly after they are addressed

codelucas:
@JMendes1995 any updates here? I am excited to integrate YAKE, but please look at my PR review first

README.rst
JMendes1995: The score refers to the relevance of the word in text.

Like says the YAKE! documentation [(),](https://github.com/LIAAD/yake#output) the lowest score means that word are more relevant.

i am not sure if i understand, you want me to change the function name to yake_keywords_ranked in file newspaper/article.py ?

Issue #234, opened 2016-04-06
Not able to parse publish date.


kartikparnami:
-> Newspaper is unable to parse date for the URL: http://pratyushsharma.blogspot.in/2016/03/jindagi-mauth-na-ban-jaye-samhalo-yaaron.html
-> On seeing page source publishing date can be seen in the line: <abbr class='published' itemprop='datePublished' title='2016-03-31T03:55:00-07:00'>3:55 AM</abbr>
->  Newspaper is able to get till this DOM element by matching it with {'attribute': 'itemprop', 'value': 'datePublished', 'content': 'datetime'} in line 203 in extractors.py
-> But the content attribute is not matched instead a new attribute title is needed. 
-> We need to add {'attribute': 'itemprop', 'value': 'datePublished', 'content': 'title'} to the PUBLISH_DATE_TAGS to get this read properly.

yprez:
@kartikparnami good find.
Are you sure it's supposed to be "title" and it's not a mistake of this specific blog?

https://schema.org/datePublished says the attribute should be "content", other sources refer to "datetime", but I couldn't find any other examples with "title"...

kartikparnami:
Well, I dont know how widespread this issue is and whether its a blog specific issue. But, I feel an addition just increases our coverage of the cases. Let me know your thoughts.

yprez:
Similar to #151

mamoit:
#402 changes the behaviour to follow the schema of datePublished.
Doesn't solve this problem in particular, but this seems to be an isolated case of out of spec metadata.

saqibaliXIQ:
not able to parse date of many domain articles
marketscreener.com
contagionlive and many other but diffbot does but its paid

Issue #151, opened 2015-06-12
Additional entry for PUBLISH_DATE_TAGS


jywsn:
I've encountered some news articles with the following metadata tags:

http://abcnews.go.com/Politics/jeb-bush-heads-europe-things-expect-trip/story?id=31635735
`<meta itemprop="datepublished" content="2015-06-09T17:02:00Z">`

http://www.wsj.com/articles/jeb-bush-challenges-russia-to-respect-its-neighbors-1433867078
`<meta itemprop="datePublished" content="2015-06-09T19:55:00.000Z">`

I would propose adding an additional entry to PUBLISH_DATE_TAGS in extractors.py, get_publishing_date():

`{'attribute': 'itemprop', 'value': 'datePublished', 'content': 'content'}`

younes0:
I've seen other date meta tags, this is a mess.
why not getting all the meta containg 'date' and retrieve the value?

younes0:
Alternative solution: retrieve all the values of meta and check if it's a date or not

mamoit:
#402 Should fix this.

PR #801, opened 2020-04-21
Move PUBLISH_DATE_TAGS to the module level for ease of modification

embray:
This would provide a simple work-around for issues like #151, #234, #402, and possibly others.  There's no good reason to define this variable in the function definition body where it cannot be accessed by users.

In general it's a better practice to define "constants" (whether or not they're mutable) such as these at the module-level or class-level.  For one, putting them in the function body actually slows down execution of the function since the list has to be rebuilt every time (this is a trivial slowdown admittedly).  But more importantly, putting them in the function definition body means they cannot easily be modified at runtime by users.

This provides a similar treatment for the language tags, as well as for the stopwords for categories (this in turn should probably be moved so that individual language modules can handle this, but that's another topic).

### DIFF
--- a/newspaper/extractors.py
+++ b/newspaper/extractors.py
@@ -75,6 +75,17 @@ PUBLISH_DATE_TAGS = [
 ]
 
 
+# This lists the tags and their associated attributes to search
+# for the article language; typically this is in one or two places
+# but you can add others to this list.  The format is similar to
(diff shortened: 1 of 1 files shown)


embray:
(fixed one small typo in the original PR)

AndyTheFactory:
PUBLISH_DATE_TAGS moved to the defines.py file in extractors

PR #815, opened 2020-05-24
Deactivate Security Certificate check

toshiro92:
During my use of newspaper, I've experienced the following error:

>  ERROR - Failed to download URL https://blogs.lentreprise.com/le-management-dans-tous-ses-etats/2019/05/13/bienveillance-empathie-demagogie-et-rupture/ due to Article `download()` failed with HTTPSConnectionPool(host='blogs.lentreprise.com', port=443): Max retries exceeded with url: /le-management-dans-tous-ses-etats/2019/05/13/bienveillance-empathie-demagogie-et-rupture/ (Caused by SSLError(CertificateError("hostname 'blogs.lentreprise.com' doesn't match either of '*.lexpress.fr', 'lexpress.fr'",),)) on URL https://blogs.lentreprise.com/le-management-dans-tous-ses-etats/2019/05/13/bienveillance-empathie-demagogie-et-rupture/.

I figured out how to quick fix it, by skipping the security certificate check by adding the `verify=False` [setting ](https://stackoverflow.com/questions/15445981/how-do-i-disable-the-security-certificate-check-in-python-requests) to the following line:
https://github.com/codelucas/newspaper/blob/9b89046d076b244fef006aee1a485b1f6e537bdf/newspaper/network.py#L62

https://github.com/codelucas/newspaper/blob/9b89046d076b244fef006aee1a485b1f6e537bdf/newspaper/network.py#L63

In my point of view, this is not necessary to check the certificate for a scraped article.

### DIFF
--- a/newspaper/network.py
+++ b/newspaper/network.py
@@ -60,7 +60,7 @@ def get_html_2XX_only(url, config=None, response=None):
         return _get_html_from_response(response, config)
 
     response = requests.get(
-        url=url,verify=False, **get_request_kwargs(timeout, useragent, proxies, headers))
+        url=url, verify=False, **get_request_kwargs(timeout, useragent, proxies, headers))
 
     html = _get_html_from_response(response, config)
 
--- a/newspaper/network.py
+++ b/newspaper/network.py
@@ -60,7 +60,7 @@ def get_html_2XX_only(url, config=None, response=None):
         return _get_html_from_response(response, config)
 
     response = requests.get(
-        url=url, **get_request_kwargs(timeout, useragent, proxies, headers))
+        url=url,verify=False, **get_request_kwargs(timeout, useragent, proxies, headers))
 
     html = _get_html_from_response(response, config)
 


codelucas:
this is a really good idea, I wonder how many requests this impacts

newspaper/network.py
```diff
@@ -60,7 +60,7 @@ def get_html_2XX_only(url, config=None, response=None):
         return _get_html_from_response(response, config)
 
     response = requests.get(
-        url=url, **get_request_kwargs(timeout, useragent, proxies, headers))
+        url=url,verify=False, **get_request_kwargs(timeout, useragent, proxies, headers))
```
codelucas, line 63: space between `, and v`

codelucas (approved):
Why does the build fail on python 3.4? Is it because of EOL for that version? or because something actually broke

I am interested in approving this so extraction success increases (what we care most about in this repo) From online discussions it seems for our purposes, since we are only reading data, never sending data, verify is not as useful for us

codelucas:
please fix my nit, then I will merge this @toshiro92

newspaper/network.py
codelucas: this is not resolved yet, please add spaces between `,` and `verify

codelucas (changes requested):
please fix my comment

newspaper/network.py
toshiro92: Done! I've commited with the requested change

toshiro92:
@codelucas Great! I agree with with you, however, I'm not sure to understand why it does not working on Python 3.4

codelucas:
I found out why it is not working on python 3.4, it is because python 3.4 is not supported anymore

toshiro92:
@codelucas Do you need another thing from me? Will you merge my fix? 😄

PR #819, opened 2020-06-25
fix broken unit test

Bachstelze:
The current build failed due to a 404 url.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -753,8 +753,7 @@ class TestDownloadPdf(unittest.TestCase):
     @print_test
     def test_article_pdf_ignoring(self):
         empty_pdf = "%PDF-"  # empty PDF constant
-        a = Article(url='http://www.technik-medien.at/ePaper_Download/'
-                        'IoT4Industry+Business_2018-10-31_2018-03.pdf',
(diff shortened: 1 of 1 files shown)


codelucas:
Great fix, I was wondering why the recent builds were having issues

tests/unit_tests.py
```diff
@@ -753,8 +753,7 @@ class TestDownloadPdf(unittest.TestCase):
     @print_test
     def test_article_pdf_ignoring(self):
         empty_pdf = "%PDF-"  # empty PDF constant
-        a = Article(url='http://www.technik-medien.at/ePaper_Download/'
-                        'IoT4Industry+Business_2018-10-31_2018-03.pdf',
+        a = Article(url='https://www.adobe.com/pdf/pdfs/ISO32000-1PublicPatentLicense.pdf',
```
codelucas, line 756: how did you find this url, ideally we should move all these test resources offline

codelucas:
done

tests/unit_tests.py
Bachstelze: It is pdf url that is already used in the unit tests. Yes, it is a bit random if the tests depend on the availability of online resources. But to a good degree the project is about this attribute.

PR #860, opened 2020-11-28
Add support for Latvian languaga

linards-kalvans:
Stopword list from https://countwordsfree.com/stopwords/latvian


bisoldi:
Looks like https://github.com/codelucas/newspaper/pull/867 is the correct PR for Latvian language support.

PR #941, opened 2022-06-12
Added ability to scrape javascript intensive apps

Sosshi:
The library was failing to scrape sites which have javascript code in it so i have added the ability to scrape such websites. So now it will be possible to scrape sites made with Vue, React and other JS intensive frameworks

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "makefile.extensionOutputFolder": "./.vscode"
+}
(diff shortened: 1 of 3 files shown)


banagale:
This sounds compelling. I noticed your change sent includes conversion of single to double quotes and some formatting.

It would be easier to review these changes if it were limited only to materially changed lines. 

While I do not believe the maintainer is approving PRs at this time, in general I'd suggest offering a PR with changes that only include what you're working on. Then consider a second that affects formatting in a more general sense.

--

All that said, I'm curious if you have test cases of sites that show article content using JS that fail using the main branch but pass using your change set.

PR #1023, opened 2026-04-01
fix: handle punkt_tab fallback in split_sentences for NLTK >= 3.9

### LINKED ISSUE #1017: WordTokenizer conflict with nltk >= 3.8.2
First, thanks for the library. Second, I found a problem with the requirements after the release of nltk == 3.8.2 for WordTokenizer

The aforementioned nltk == 3.8.2 solves a remote code execution vulnerability, disabled to obtain "punkt" and replaced it with "punkt_tab".
Issues:
- https://github.com/nltk/nltk/issues/3266#issuecomment-2284001819
- https://github.com/nltk/nltk/issues/3293

This creates a problem with newspaper3k when downloading "punkt" in REQUIERED_CORPORA. Then I suggest modify REQUIRED_CORPORA or updating the requierements.py to avoid using nltk >= 3.8.2

danishashko:
Fixes #1017

NLTK 3.9 moved the default punkt tokenizer from `tokenizers/punkt/english.pickle` to `tokenizers/punkt_tab/english`. On a fresh install with NLTK >= 3.9 only `punkt_tab` is downloaded, so calling `split_sentences()` raises a `LookupError` even when NLTK is properly installed.

**What this changes:**

1. `newspaper/nlp.py` - `split_sentences()` now tries the old path first and falls back to `punkt_tab` if not found. Works with both old and new NLTK versions.
2. `download_corpora.py` - adds `punkt_tab` to `REQUIRED_CORPORA` so it gets downloaded alongside `punkt`.

PR #1006 already adds `punkt_tab` to the download list, but the runtime crash in `nlp.py` still happens if only `punkt_tab` is available. This fixes both.

### DIFF
--- a/download_corpora.py
+++ b/download_corpora.py
@@ -8,6 +8,7 @@ import nltk
 REQUIRED_CORPORA = [
     'brown',  # Required for FastNPExtractor
     'punkt',  # Required for WordTokenizer
+    'punkt_tab',  # Required for WordTokenizer (NLTK >= 3.9)
     'maxent_treebank_pos_tagger',  # Required for NLTKTagger
     'movie_reviews',  # Required for NaiveBayesAnalyzer
     'wordnet',  # Required for lemmatization and Wordnet
--- a/newspaper/nlp.py
+++ b/newspaper/nlp.py
@@ -154,7 +154,10 @@ def split_sentences(text):
     """Split a large string into sentences
     """
     import nltk.data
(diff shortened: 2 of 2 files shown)


pgwimberl71 (commented):
1f6e1770c95468d862d86084d290aa8b0e8b5db71f6e1770c95468d862d86084d290aa8b0e8b5db7

PR #1028, opened 2026-08-23
fix(date): recover publication dates the URL and JSON-LD already state

idoshamun:
## What

`get_publishing_date()` was wrong in **both** directions, and both faults trace to one line — it scanned the whole URL with `STRICT_DATE_REGEX`, which captures the separators around its match.

### It missed real dates

A `/2014/04/` path arrives as `2014/04/`, and `dateutil` raises on the trailing slash. Every year-month URL fell through to the metadata strategy, and a page without date metadata then ended up with **no date at all**.

The in-code comment blamed *"URL dates without a day specifier"*, which sent everyone looking in the wrong place:

```
parse('2014/04')   -> 2014-04-23    # fine
parse('2014/04/')  -> ParserError   # the actual failure
```

Replayed over 20 real year-month URLs from our corpus, the old code parsed **0 of 20**.

### It invented dates that were not there

The worse direction. `kali-linux-2026-1-release` and a Shopify `...-2026-07-...` API version both look like dates to a regex that scans anywhere in a string. They only ever came out right *by accident* — the same trailing separator that hid the real dates also hid them. Naively fixing the strip would have turned five accidental-correct nulls into five confident wrong dates.

**The discriminator is position, not shape.** A date in a URL owns the start of its path segment (`/2013/03/slug`, `/2025-10-registry-directory`); a version buried in a slug does not. `_get_url_date` walks path segments instead of scanning.

| URL | before | after |
|---|---|---|
| `firebase.blog/posts/2013/03/…` | — | 2013-03-01 |
| `mypy-lang.blogspot.com/2024/10/…` | — | 2024-10-01 |
| `ui.shadcn.com/docs/changelog/2025-10-registry-directory` | — | 2025-10-01 |
| `github.blog/changelog/2026-08-21-the-new-…` | — | 2026-08-21 |
| `techcrunch.com/2026/08/21/…` | 2026-08-21 | 2026-08-21 |
| `kali.org/blog/kali-linux-2026-1-release` | — | — |
| `shopify.dev/changelog/…-2026-07-…` | — | — |

## Metadata extraction

Adds schema.org **JSON-LD `datePublished`** (bare object, list, and `@graph` shapes; malformed JSON can never propagate), `<time>` elements the document *marks* as the publication date, and three more meta names (`date`, `dc.date`, `dcterms.created`).

Over 22 sampled pages that plainly state a date, the existing 11-tag list found **6**. JSON-LD is where most current site generators put it. A bare `<time>` is deliberately ignored — it is as likely to be a reading time or a comment stamp.

Both run **after** the existing tags, so no page that resolves a date today changes its answer. There's a test pinning that.

## Determinism

Partial dates are now anchored to an explicit `default`. Given `2014/04`, `dateutil` fills the missing day from **today**, so the same page yielded a different date depending on when it was parsed and a re-extraction silently moved the answer. A year-month now means the 1st.

## Why this matters downstream

daily.dev's claim ledger holds **11,080 claims dated from the day a page was crawled rather than published** — a page with no recoverable date leaves every consumer falling back to crawl time. That is harmless on a live feed and wrong by the article's whole age on an archive import (2022 articles drift four years). Two of the three ways a date goes missing are fixed here.

Related: dailydotdev/yggdrasil#725 (carry the feed's `<pubDate>` through, for pages that state no date anywhere) and dailydotdev/daily-api#4139 (stop the ledger dating a claim from crawl time).

## Tests

7 new tests covering both URL directions, the day-pinning, the three JSON-LD shapes, malformed JSON, `<time>`, and meta precedence. The suite goes 35 → 42 passing; the 8 pre-existing failures (missing nltk corpora, network-dependent image tests) are identical before and after.

## Note for whoever merges

`pyproject.toml` pins newspaper at `4e8b3f46` while `requirements.txt` pins `46fed320`, and the README says CI/CD uses `requirements.txt`. Those two need to move together, or the deploy will not pick this up.

### DIFF
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -213,18 +213,21 @@ class Article(object):
         self.throw_if_not_downloaded_verbose()
 
         self.doc = self.config.get_parser().fromstring(self.html)
-        self.clean_doc = copy.deepcopy(self.doc)
 
         if self.doc is None:
             # `parse` call failed, return nothing
             return
 
+        document_cleaner = DocumentCleaner(self.config)
+        output_formatter = OutputFormatter(self.config)
+
+        self.clean_doc = copy.deepcopy(self.doc)
+        # Before any computations on the body, clean DOM object
+        self.clean_doc = document_cleaner.clean(self.clean_doc)
+
         # TODO: Fix this, sync in our fix_url() method
         parse_candidate = self.get_parse_candidate()
         self.link_hash = parse_candidate.link_hash  # MD5
-
-        document_cleaner = DocumentCleaner(self.config)
-        output_formatter = OutputFormatter(self.config)
 
         title = self.extractor.get_title(self.clean_doc)
         self.set_title(title)
@@ -266,9 +269,6 @@ class Article(object):
         self.publish_date = self.extractor.get_publishing_date(
             self.url,
             self.clean_doc)
-
-        # Before any computations on the body, clean DOM object
-        self.doc = document_cleaner.clean(self.doc)
 
         self.top_node = self.extractor.calculate_best_node(self.doc)
         if self.top_node is not None:
--- a/newspaper/article.py
+++ b/newspaper/article.py
@@ -275,8 +275,8 @@ class Article(object):
             video_extractor = VideoExtractor(self.config, self.top_node)
             self.set_movies(video_extractor.get_videos())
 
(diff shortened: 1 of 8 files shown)


idoshamun:
Opened against the wrong remote by mistake — this belongs on our dailydotdev fork. Sorry for the noise.

